PR #46, opened 2017-10-29
Singleplayer fixes

ShadowTeddy:
Added fixes for Hit-fix crashing singleplayer, Hit-Fix should only run within the multiplayer instance of Halo 2 (And shouldn't interfere with SinglePlayer).

-Used recent release of cartographer/development
-Added support for 60fps cutscenes.

### DIFF
--- a/xlive/H2MOD.cpp
+++ b/xlive/H2MOD.cpp
@@ -1007,31 +1007,33 @@ int __cdecl OnMapLoad(int a1)
 	}
 	int ret = pmap_initialize(a1);
 
+	if (*GameEngine == 2)
+	{
 #pragma region Apply Hitfix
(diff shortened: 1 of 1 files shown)


iSupersniper (dismissed):
Your change in H2MOD.cpp on line 1277 should not have been done.  This only will check team values greater than 1.  Observation team is 255 or in decimal -1.  This will break RvB mode. 

Please revert the change and resubmit the commit.

iSupersniper:
Your change in H2MOD.cpp on line 1277 should not have been done. This only will check team values greater than 1. Observation team is 255 or in decimal -1. This will break RvB mode.

Please revert the change and resubmit the commit.

PR #48, opened 2017-11-06
Halo 2 Xbox Rebalance

ShadowTeddy:
We've gotten most weapons re-balanced to H2X's behavior, its now just a matter of testing it.
Lemme know if any issues occur.

Variant triggers:
_"H2X" OR "h2x"_

Thanks again to UncutFighter and Hootspa for their collaboration, this would not have been possible without their input. (And RE4thewin for the carbine)


iSupersniper:
Before the code can be pushed Please compile a DLL and allow for testing.

PR #57, opened 2017-11-24
H2X gametype

nukeulater:
"This includes everything from the previous pull request & some new additions.

H2X Rebalance:
We've gotten most weapons re-balanced to H2X's behavior, its now just a matter of testing it.
Lemme know if any issues occur.

Variant triggers:
"H2X" OR "h2x"

Thanks again to UncutFighter and Hootspa for their collaboration, this would not have been possible without their input. (And RE4thewin for the carbine)

And raw mouse input.
Thanks to hootspa for collaborating on raw mouse input, we finally did it!"

Added discord_enable and raw_input to xlive.ini in order to enable/disable those.

### DIFF
--- a/xlive/H2MOD.cpp
+++ b/xlive/H2MOD.cpp
@@ -1654,12 +1654,12 @@ void H2MOD::Initialize()
 	h2mod->ApplyHooks();
 
 	extern UINT discord_enable;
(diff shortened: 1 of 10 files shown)


shayaantx:
yours has conflicts, resolve and resubmit

PR #76, opened 2017-12-07
Cleaned H2X code

nukeulater:
- added maxplayers command argument validation

### DIFF
--- a/xlive/H2MOD_H2X.cpp
+++ b/xlive/H2MOD_H2X.cpp
@@ -14,7 +14,7 @@ DWORD getOffset() {
 }
 
 
-void H2X_Initialize()
+void H2X::Initialize()
 {
 	
 	DWORD FloatOffsets = getOffset();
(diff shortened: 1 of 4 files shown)


shayaantx:
1) you instantiated cstr in console commands as a array, so your corresponding delete needs to reflect that

instead of delete cstr; delete[] cstr;

2) void H2X_Initialize(bool init);

This is ugly, this is a boolean trap, and frankly passing a boolean named init into a function named initialize makes no sense. From reading your code change in the init method, you are effectively resetting values if init is true or not. Just make a method that resets the values or call deinit for the game variant.

Also, git gud

xlive/H2MOD_H2X.cpp
```diff
@@ -1,48 +1,52 @@
...
+	int offset = 0x47CD54;
+	if (h2mod->Server)
+		offset = 0x4A29BC;
+	TRACE_GAME("[h2mod] H2X is being run on client game");
+	DWORD FloatOffsets = *(DWORD*)((char*)h2mod->GetBase() + offset);
+	return FloatOffsets;
+}
+
+
+void H2X_Initialize()
```
num0005: make it a static class member instead.

shayaantx:
I really should make you make constants in h2x variant, but ill let it slide

PR #77, opened 2017-12-07
Redirect print in scripts calls to our console.

num0005:
They say a picture is worth a thousand works so here is one.
image


shayaantx:
looks good to me

PR #84, opened 2017-12-26
Added sensitivity to config file

nukeulater:
*Changed setSens function a little

### DIFF
--- a/xlive/H2Config.cpp
+++ b/xlive/H2Config.cpp
@@ -80,6 +80,8 @@ bool H2Config_controller_aim_assist = true;
 int H2Config_fps_limit = 60;
 int H2Config_field_of_view = 0;
 float H2Config_crosshair_offset = NAN;
+int H2Config_sens_controller = 0;
+int H2Config_sens_mouse = 0;
 bool H2Config_disable_ingame_keyboard = false;
 bool H2Config_hide_ingame_chat = false;
 int H2Config_custom_resolution_x = 0;
@@ -205,6 +207,12 @@ void SaveH2Config() {
 			fputs("\n# 0 - In-game chat is displayed normally.", fileConfig);
 			fputs("\n# 1 - In-game chat is hidden.", fileConfig);
 			fputs("\n\n", fileConfig);
+			fputs("# controller_sensitivity Option (Client):", fileConfig);
+			fputs("\n# <value> Change controller sesitivity to your preference.", fileConfig);
+			fputs("\n\n", fileConfig);
+			fputs("# mouse_sensitivity Option (Client):", fileConfig);
+			fputs("\n# <value> Change mouse sesitivity to your preference.", fileConfig);
+			fputs("\n\n", fileConfig);
 			//fputs("# custom_resolution Options (Client):", fileConfig);
 			//fputs("\n# <width>x<height> - Sets the resolution of the game via the Windows Registry.", fileConfig);
 			//fputs("\n# 0x0, 0x?, ?x0 - these do not do modify anything where ? is >= 0.", fileConfig);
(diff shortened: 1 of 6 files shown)


Himanshu-01 (commented):
Nuke you are still doing it the wrong way here

https://github.com/PermaNulled/cartographer/pull/84/files#diff-b6c2d0fbf3f50a7493e3eda1e80e0d0dR230

Change it to 
```DWORD maxPlayersNumber =lobby_globals + 0x4C80;```

LOOKING AT the way u set it in later place 
```*(BYTE*)maxPlayersNumber = maxPlayersSet;```
The variable should hold address which cannot be stored in A BYTE

File: xlive/H2Config.cpp [hunk 1]
```diff
@@ -686,6 +705,30 @@ static int interpretConfigSetting(char* fileLine, int version, int lineNumber) {
 				est_hide_ingame_chat = true;
 			}
 		}
+		else if (!H2IsDediServer && sscanf(fileLine, "controller_sensitivity =%d", &tempint1) == 1) {
+			if (est_sens_controller) {
+				duplicated = true;
+			}
+			else if (!(tempint1 >= 0)) {
```
shayaantx, line 712: all these if/elseif/else seem like they can be one liners

also comments would be nice

xlive/H2ConsoleCommands.cpp
```diff
@@ -230,7 +230,9 @@ void ConsoleCommands::handle_command(std::string command) {
 			BYTE maxPlayersNumber = *(BYTE*)(lobby_globals + 0x4C80);
 
 			if (isNum(cstr)) {
-				int maxPlayersSet = atoi(cstr);
+				delete[] cstr;
```
shayaantx, line 233: you should use a try finally here, in the finally block you can do the delete in there and outside the try do char *cstr = NULL;
inside the try, do cstr = new char[firstArg.length() + 1];

then in finally, do if (cstr != null) delete[] cstr;

xlive/H2MOD.cpp
```diff
@@ -973,9 +973,10 @@ bool __cdecl OnPlayerSpawn(int a1)
 
 	int PlayerIndex = a1 & 0x000FFFF; 
 
-	if (b_Infection)
+	if (b_Infection) {
```
shayaantx, line 976: the way i read the old code, we called spawnPlayer regardless of this if block, are there any edge cases here where spawn player does not get called?

xlive/H2Tweaks.cpp
```diff
@@ -197,12 +197,17 @@ void DeinitH2Tweaks() {
 
 }
 
-void setSens(short input_type, float sens) {
+void setSens(short input_type, int sens) {
+	if (sens == 0)
```
shayaantx, line 201: you should doc what sens == 0 really means and why we return out

shayaantx (changes requested):
review

shayaantx:
and address whatever himanshoe said

xlive/H2MOD.cpp
nukeulater: no, i changed because the second SpawnPlayer func from inf class was getting called even though infection wasn't enabled, line 978

xlive/H2Config.cpp
nukeulater: That's up to Glitchy to change

KILLER-CHIEF: Please explain how that can be simplified without losing current functionality.
Also those lines are self documenting. The only comments I'll ever add is a description for what that entire block of code does where the chain of if-else-ifs begin.

shayaantx: you have a boolean called "duplicated" wtf does that even mean? Thats not self documenting, thats a poorly named bool

then TempInt1, another poorly named boolean.

This is not self documenting, this is only self documenting after you look around and figure out whatever everything means and where every value comes from.

shayaantx: I don't care much about breaking the if/elseif/else into one liners, more so into starting to make the code more readable to anyone.

PR #89, opened 2018-01-08
Allow maps with custom huds and tags on dedis too

nukeulater:
Changed the bytes count because only 7 are needed for the hook.


KILLER-CHIEF:
Yo, like you obviously discovered after i mentioned it was weird no login error messages were being displayed on the dedicated server, you really need to ask why thing are they way they are written before you go changing other peoples code. I trust the modified client side hook still functions. And hope that the dedicated server hook also functions correctly. We will be sure to test this in dev preview though as well.

PR #94, opened 2018-01-14
Apply Hitfix in MP game engine only

bigtweekx:
Again... I notice that the game engine check this was removed, was this intentional?

### DIFF
--- a/xlive/H2MOD.cpp
+++ b/xlive/H2MOD.cpp
@@ -836,28 +836,31 @@ int __cdecl OnMapLoad(int a1)
 	
 #pragma region Apply Hitfix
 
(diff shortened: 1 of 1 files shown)


ShadowTeddy (commented):
If you have a look at the line of code starting at 811 (H2MOD.cpp), this is already present so that if statement doesn't need to be re-added. It would be useless code at that point. 

You can tell if this function is working by loading a campaign mission up on 06a_sentinelwalls (Sacred Icon) and watching the cut-scene, the scene where 343 Guilty Spark comes into frame is typically when the game used to crash.

bigtweekx:
didnt notice that thanks

PR #99, opened 2018-01-25
BlamLibrary Addon

Himanshu-01:
### Description ###
Adds a list of BlamLibrary Support for Project Cartographer
Its still initial stage and lots of things need to be added but the definitions can still be put into very much use
### Why This should be merged ###
Properly arranges the game definitions into structs and enums and makes the code more readable

### Issues ###
Havent found any issues beside the incomplete structures that need to be still done

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,3 +31,4 @@ xliveless.v12.suo
 /Halo2
 /xlive/Debug
 /xlive/Release
+*.opensdf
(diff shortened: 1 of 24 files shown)


.gitignore
```diff
@@ -31,3 +31,4 @@ xliveless.v12.suo
 /Halo2
 /xlive/Debug
 /xlive/Release
+*.opensdf
```
shayaantx, line 34: why is this being added? please remove unless necessary

shayaantx (approved):

.gitignore
Himanshu-01: cuz its  a practically useless file in my regard :)

PR #107, opened 2018-02-18
Added on-screen FPS counter

nukeulater:
Don't run H2X on dedicated server because it's useless (players can change it on the fly and still sync, so no need of it on dedis).
Run discord rich presence only on instance 1 of the game.
Got rid of some unused variables.

https://i.imgur.com/qlFah5q.png


shayaantx:
why are you adding a fps counter?

PR #114, opened 2018-03-23
Proper 60fps cutscenes fix

nukeulater:
With the hook, we most likely break the game's functionality by always returning 0, so patch the function only where needed.
Run H2X only in multiplayer and don't apply on dedicated server because the Firerate code is client side.
Run discord only on instance 1 of the game to avoid any problems that may occur.
Compiler warning fix.
Typos fix.
Added missing include guard to header file.
MouseInput Initialize now static class member.
Replaced OnMapLoad with onGameEngineChange.
Rebased with Development.
Reworked FOV code a little (to scale properly the third person camera FOV with FP).
Fixed elite_mp lights/orange glow. (Thanks Himanshu for help)

### DIFF
--- a/xlive/Detour.h
+++ b/xlive/Detour.h
@@ -1,1 +1,2 @@
+#pragma once
 extern void Detour();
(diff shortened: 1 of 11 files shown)


shayaantx:
who has tested this

nukeulater:
I did test these changes. in the current release, if framerate is uncapped the cutscenes will run in slow motion, so i fixed that. Discord will now run only on instance 1 of the game so we avoid any issues that might appear. H2x is running in single player and thats not a good thing to do if we dont want to break coop further more. Also the code related to fire rates is client side so no need of running h2x on dedi.

shayaantx (approved):

PR #117, opened 2018-03-30
Remove encryption and add interface to the games internal file system api.

num0005:
Files that are currently encrypted are left as is and replaced with an unencrypted version when they written to.


xlive/Blam/Engine/FileSystem/FiloInterface.cpp
```diff
@@ -0,0 +1,89 @@
+#include "FiloInterface.h"
+#include "../H2MOD.h"
+
+namespace FiloInterface
```
shayaantx, line 4: fileinterface?

shayaantx (approved):

PR #124, opened 2018-05-22
Fix crash caused by renaming the executable.

num0005:
The code for detecting if xlive got loaded into a server or a client now has a fallback that checks the original filename of the executable and will show an error instead of crashing if that fails.


shayaantx:
who tested this

PR #125, opened 2018-05-27
Changed offsets again for onGameEngineChange..

nukeulater:
To fix AI not spawning in MP.
Hopefully this is the last time changing this.

### DIFF
--- a/xlive/H2MOD.cpp
+++ b/xlive/H2MOD.cpp
@@ -660,10 +660,10 @@ bool __cdecl player_remove_packet_handler(void *packet, int size, void *data)
 }
 */
 
-typedef int(__cdecl *originalFunc)();
+typedef void(__cdecl *originalFunc)(int a1);
 originalFunc p_originalFunc;
 
-int __cdecl onGameEngineChange()
+void __cdecl onGameEngineChange(int a1)
 {
 	overrideUnicodeMessage = false;
 	isLobby = true;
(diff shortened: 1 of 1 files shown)


xlive/H2MOD.cpp
```diff
@@ -1118,8 +1119,8 @@ void H2MOD::ApplyHooks() {
 		pspawn_player = (spawn_player)DetourFunc((BYTE*)this->GetBase() + 0x55952, (BYTE*)OnPlayerSpawn, 6);
 		VirtualProtect(pspawn_player, 4, PAGE_EXECUTE_READWRITE, &dwBack);
 
-		PatchCall(GetBase() + 0x86C3, onGameEngineChange);
-		p_originalFunc = (originalFunc)(GetBase() + 0x48CA5);
+		PatchCall(GetBase() + 0x49E95, onGameEngineChange);
```
shayaantx, line 1122: why does this keep changing?

shayaantx: why did you also change the original code I had that did this

nukeulater: Well, the first hook i did was causing crashes if player was restarting the mission, and the second code change was causing AI not to spawn in MP, because all the AI initialization was done before this method, so now I change it again so both issues are gone. I promise this is the last change I do to this.

PR #126, opened 2018-05-29
Enable live menus only if person has Dev account.

nukeulater:
To prevent people from accessing the live servers until they are ready.


shayaantx:
can u attach a screenshot showing enabled/disabled

nukeulater:
Disabled:
disabled

Enabled:
enabled

PR #146, opened 2018-10-10
Added offset to deserialize membership packet for dedi 

nukeulater:
Fixes crash.


nukeulater:
I didn't check but all the network hooks use the same size for both client and server. Also i did test if it works and it does work.

PR #192, opened 2019-01-03
Revert window name change to fix compatibility with H2V Resizer

iSupersniper:


iSupersniper:
Not needed right now.

PR #194, opened 2019-01-12
Disable hitmarker sound effect and add a voice chat option.

General-101:
Same as before but like now it has a voice chat thing. I don't know how to Github.


iSupersniper:
You should properly handle the voice chat initialization and  deinitialization with the gui option rather than having players restart their games.

PR #204, opened 2019-03-17
Re-enable hitfix on sniper and gauss.

nukeulater:


iSupersniper:
Can you also change the battle rifle back to 1200, and the gauss back to 200 as well?

iSupersniper:
Code looks fine, please do a pull request to the development branch so we can keep release branches separate.  Thank you :)

PR #212, opened 2019-04-13
Faster Live server list downloading + misc changes

General-101:
-Force Refresh Rate
Saves and forces a selected refresh rate from the Carto settings menu on startup
-Save mouse/controller sens
Saves the value last inputed into the console and pokes it on map load. The default value is zero and the code does nothing if it is zero.
-Fix for console spawn commands
Should only be able to be used by the host now
-Hitfix reverted to 4.0.4 values
-Live server list downloads faster
Original commit by Nuke found here
NukeULater@1d14f0d
-Fix for Xinput copying
This fixes splitscreen. Original commit by num005 found here
num0005@2e761c4

num0005@d23e8f0

num0005@82c6455

num0005@09be95a

-Commented out some unused features

### DIFF
--- a/xlive/H2MOD.cpp
+++ b/xlive/H2MOD.cpp
@@ -1129,6 +1129,7 @@ void __cdecl OnMapLoad(int a1)
 		H2Tweaks::disableAI_MP();
 		UIRankPatch();
 		H2Tweaks::disable60FPSCutscenes();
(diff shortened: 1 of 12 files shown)


iSupersniper:
Everything looks good.
We need to go in and document the code for future updates.

PR #217, opened 2019-05-11
Temp solution to OGH2 and H2X flaw

iSupersniper:
patch to fix ogh2 and h2x from running at the same time.


iSupersniper:
In favor of pull request https://github.com/PermaNulled/cartographer/pull/219

PR #218, opened 2019-05-11
OGH2 & H2X Conflict Fix + Case Insensitive Comparison

atticuslaurs:


atticuslaurs:
I built the DLL with this change -- tested it and it works as expected -- if you have h2x & OGH2 in a game variant name, then it uses h2x not ogh2. If you select OGH2 then it uses OGH2

atticuslaurs:
oops this is what you get when you make the change rearrange files and don't attempt to rebuild to check lol brb

PR #219, opened 2019-05-12
Patch H2X & OGH2 Conflict + Case Ins. Comp. for most game variant mods

atticuslaurs:
Alternative way to patch the OGH2 & H2X conflict with simple if else structure; and case insensitive comparison with StrStrIW() replacing calls to wcsstr() @ PermaNull's Suggestion


atticuslaurs:
I've Tested this build a little to verify that you can apply the mods as expected and with whatever case you wish.

iSupersniper:
I appreciate the interest in coding for this project.

StrStrI is indeed proper for scanning for non case sensitive strings. However it does so by creating a copy of the string in memory and then referencing it as the defined string.  This makes the process a little slower and causes more memory to be used.

Instead of doing it that way, using wcsstr is a more precise search function that allows for a specific string to be found.  Yes we have to define more cases (due to it being case sensitive) but it provide a less likeliness of messing up. 

of course this all depends on user preference and the type of project being created.  Because we are reading game run-time it is best to make it as precise and short as possible to prevent any cause of game run-time delay.

atticuslaurs:
I should mention it was PermaNull who suggested the function. So I decided to try it out & it works as expected. It makes those really long trailing lines nicer to read too ; )
However I am aware that there is a performance cost.

iSupersniper:
@PermaNulled can you clarify?  Because to my understand it seems like a slower method.

pnill:
Considering the function only gets called when starting a map, the additional memory usage shouldn't be a concern and gives us cleaner code. I'd say that trade off is an acceptable one.

I should also mention that it clears the memory after creating the copy if that is actually what it does... I don't see a reason for this function to do that outside of possibly needing to manipulate the string.

atticuslaurs:
: 3

shayaantx:
there would be no notable performance problem, this method would be fine (in debug mode it might be noticed if you are profiling the heap, since free_dbg leaves around blocks of free-d memory, and don't know what all these garbage blocks of data are)

however I'd avoid the extra import, use "_wcsicmp" instead. I'd also make the variant consts defined once at the top of H2MOD

atticuslaurs:
Thanks for the suggestions I'll do that when I have a chance. Currently testing out Nuke's Mouse-troller fix

atticuslaurs:
Hm well if I understand that function "_wcsicmp" correctly it seems to do direct comparison of strings determining if they are equal rather than finding if the sub-string is present in the overall string. Checking for validity is a little more confusing since it returns whether or not the strings are greater than or less than each other, and the gametype may have more than just "h2x" in it for example. If we were going to go that route I think it would be easier to just convert the variant_name to all lower case characters and then just use wcsstr() to see if the mods are in the name.

atticuslaurs:
Yea I tested and it seems to just look for equality rather than testing for substring so if I add anything else into the name the mod isn't applied

pnill:
That's because he provided a string compare not a string search like strstr.

atticuslaurs:
Well in any case I've committed the string constants--but kept StrStrIW

iSupersniper:
It passes all the checks, but before I commit to the dev branch please compile a local dll and let the dev preview team test each variant with different case letters to verify it works properly and all game modes are initiated properly.

Also try to note any increased loading times or stuttering just in case.  Also when you compile the local dll, make sure it is in RELEASE and not DEBUG.  It makes a difference.

atticuslaurs:
Okay

atticuslaurs:
Will do; don't have perm to post to #test_dll yet tho, so I'll just put it in #dev_preview

atticuslaurs:
ok uploaded

PR #222, opened 2019-05-29
Big fixes

almic:
I tried my best to make each commit individually independent, so if you don't want to merge a specific commit let me know and it wouldn't be much work to keep everything else.

There is a nice tl;dr at the bottom, that literally covers everything in like two sentences.

Okay here goes...

## Minor Additions
- Discord source files
- Libcurl source files
- Miniupnpc source files
- Opus source files
- Portaudio source files
- Google protobuf compiler executable
- Batch file to use protobuf.exe on local terminal path (see localprotobuf.bat for more info)

## Minor Deletes (deletion? removals?)
- Removed a lot of unneeded google protobuf code, the entire compiler source code was included. Instead, just the relevant code used by compiled protobuf programs are left, and an executable now replaces the compiler source code.
- LIB files are totally gone, they are built along with xlive.dll instead now (part of premake commit, see below)
- VXCProject files are totally gone, they are generated using premake (see below)

## Major additions
- Premake is now used to generate project files, this reduces repository size and localizes build options, commands, defines, dependencies, external libraries, build configurations, etc. into only a few easily editable .lua files

## Minor changes
- Project now defines `stdafx.h` as a precompiled header, which was sporadically used already.
- Some include paths have shifted, but otherwise are identical while programming. This change was made because it follows the standard dependency inclusion file structure of "dependency/include/dep_name/[source files]". This forces include paths to better specify the third party dependency name, while maintaining relatively unchanged source code of those dependencies and existing include paths

## NO MORE WARNINGS AT ALL
There are absolutely zero warnings when building, I only cheesed this once with a single `#pragma warning` tag, but there was simply no alternative. The project no longer hard ignores warnings in the settings. The `Big Stuff` commit contains all these warning fixes. Mostly just harmless compiler warnings about value truncation and unexpected boolean conversions.

## FAQ
**Why in gosh darn diddly should we pull this?**

This new system is going to be much easier to develop with, and is much friendlier to newbies of the project. It also forces a standard of zero warnings, something that make building much easier and promotes much cleaner code.

**I can't pull these changes because of conflicts!**

None of the original files have been changed significantly. The best thing to do is copy just the files you've changed to some other directory, re-clone the repository, and copy+paste your changed files back in. You may get some warnings about a missing include for `stdafx.h` or `google/protobuf/...` can't be found. Just add the PCH include where it asks, and remove the `google` part from protobuf paths. That should fix everything. If you get warnings, check the original file source where you get the warnings and copy the fix.

## TL;DR
No more warnings, we build all the dependencies instead of using prebuilt libs, and premake is used to generate project files. As far as you can be concerned, literally nothing has really changed except `google/protobuf` includes need to have the "google" bit removed.


almic:
Just discovered that appveyor expects visual studio project files. This should be no big deal to fix but I can't do it on my end.

FAQ because I know those two questions are going to be asked, especially the conflicts question, much easier that way than doing what GIT would suggest by stashing changes/ resolving conflicts and such

almic:
I'm closing this, going to come at this from a different angle with submodules.

PR #223, opened 2019-06-02
Dev logging

almic:
Yeah this is basically just sprinkles it doesn't really fix anything, well maybe it fixes one thing in the wrong way.

### Minor misc
Felt like mentioning this right away, this makes use of the PCH `stdafx.h` now, so some file changes only add it to the header includes. Some new inclusions have been added to the PCH as well.

### New 3rd party lib: spdlog
This is a thread-safe logger with a lot of juicy stuff, see the repo [here](https://github.com/gabime/spdlog)

### Logging implementation
- Changed the name scheme for log defines: `TRACE_GAME` --> `LOG_TRACE_GAME`
- Generic logging that uses `LOG_TRACE` goes only to console output now (see below about config)
- OnscreenDebug uses the new logging system instead of it's own implementation (still does everything the same, just with the new stuff)
- log.cpp created to hold some function definitions, so implementation could be changed without having to rebuild the PCH every time
- log.h included in PCH, now every file can access logging, and is in fact forced to
- Colors in the console are replaced with level tags in file outputs
- Date is stripped from console output, but retained in file output
- Console prints all logging from everywhere, and uses an extended naming pattern to differentiate between loggers in the window, which would not be retained in file output (see future suggestions below)
- Game should always start no matter if log files fail to be created, although a single window alert will appear (ONLY ONCE) if any log files cannot be created.
- New formatting style, see existing formatted logs or the spdlog repo for more information about it.

### Configuration
- Console, just like debug logging, is disabled by default on new installations.
- Console window has it's own config setting named `debug_log_console`
- Log level can be changed with new config setting `debug_log_level`, which defaults to 2 (info)
- Log files attempt to first be created in the `AppData\Local\Microsoft\Halo 2\logs` directory, then in the local install directory folder `logs`, and then just inside the game directory if all that fails.

### Future suggestions
- A single file sink could be created that collects all logs, while retaining the individual log files.
- Update existing log calls to use appropriate levels, currently 99% of all logs are `TRACE` level.
- Save console logging to file system, right now console-specific logs are console only.


iSupersniper:
No issues posted in dev preview, therefore I will merge this to continue with other project merges.

iSupersniper:
No issues posted in dev preview, therefore I will merge this to continue with other project merges.

PR #230, opened 2019-06-27
fix file flush

almic:
File output works much better, although it may slow down the game during extremely high logging, which for now does not happen in the release version.

OnScreenDebug logger moved to Startup, so it can be properly destroyed on game exit.

Due to the order of config file loading and OnScreenDebug initialization, the log-level for OnScreenDebug will always be `trace`, although it only uses `debug` level for now. This was preventing file logging from working because the default level is `info`, and is not updated according to the config file until after OnScreenDebug loads. This should be fine, as long as OnScreenDebug isn't abused in the future.


num0005 (approved):

xlive/H2MOD/Modules/OnScreenDebug/OnscreenDebug.cpp
```diff
@@ -49,7 +48,7 @@ void addDebugText(char* text) {
 	strncpy(DebugStr[DebugTextArrayPos], text, lenInput);
 	memset(DebugStr[DebugTextArrayPos] + lenInput, 0, 1);
 
-	onscreendebuglog->debug(text);
+	onscreendebug_log->debug(text);
```
shayaantx, line 51: no underscores, camelcase for variable names (yes I know its not consistent across the codebase)

xlive/H2MOD/Modules/Startup/Startup.cpp
```diff
@@ -37,6 +37,9 @@ h2log *network_log = nullptr;
 // Map loading logger
 h2log *checksum_log = nullptr;
 
+// OnScreenDebug logger
```
shayaantx, line 40: this doc seems unnecessary, the code is self documenting

xlive/Util/log.cpp
```diff
@@ -19,6 +19,9 @@ h2log::h2log(const std::string& name)
 {
 	sname = name;
 	wname = string_convert.from_bytes(name);
+
+	// should be fine to do this for every logger, better here once than someplace else
```
shayaantx, line 23: second part of the doc seems unnecessary

if you actually have a performance concern calling flush everytime a log is constructed then I'd question this entire line. It really depends how often we are constructing log objects, if its not often, the doc isn't necessary.

nukeulater (approved):

PR #236, opened 2019-07-07
Fix more shaders broken on Nvidia.

num0005:
Fixes the vines issue on Ivory Tower among other issues that are caused by that template.


num0005:
Just refactors the code to null another bitmap that's also broken on Nvidia.

PR #242, opened 2019-07-14
Remove socket from sockmap after XCloseSocket call

nukeulater:
Don't play default sounds if custom sounds are not present.
Added in-game selection support for 4k, 1440p and 480p. (will probably not fix the text dissapearing  for high def resolutions)
Some code clean-up.

**IT CAN BE MERGED**

 Fixed clients not being able to join dedicated servers hosted on the same computer, but the other peers are getting kicked out if local clients are in.
But still, the changes are valid.

### DIFF
--- a/xlive/H2MOD.cpp
+++ b/xlive/H2MOD.cpp
@@ -14,7 +14,6 @@
 #include "H2MOD\Variants\GunGame\GunGame.h"
 #include "H2MOD\Modules\Networking\Memory\bitstream.h"
 #include "H2MOD\Modules\Networking\CustomPackets\CustomPackets.h"
-#include <Shlwapi.h>
 #include "H2MOD\Tags\global_tags_interface.h"
 #include "MetaLoader\tag_loader.h"
 
@@ -881,7 +880,7 @@ void H2MOD::CustomSoundPlay(const wchar_t* soundName, int delay)
 		if (delay > 0)
 			std::this_thread::sleep_until(timePoint);
 
-		PlaySound(soundName, NULL, SND_FILENAME);
+		PlaySound(soundName, NULL, SND_FILENAME | SND_NODEFAULT);
 	};
 
 	if (!h2mod->Server)
--- a/xlive/H2MOD.cpp
+++ b/xlive/H2MOD.cpp
@@ -17,15 +17,19 @@
 #include "H2MOD\Tags\global_tags_interface.h"
 #include "MetaLoader\tag_loader.h"
 
-H2MOD *h2mod = new H2MOD();
+H2MOD* h2mod = new H2MOD();
 GunGame* gunGame = new GunGame();
-Halo2Final *h2f = new Halo2Final();
+Halo2Final* h2f = new Halo2Final();
+DeviceShop* device_shop = new DeviceShop();
 XboxTick* xboxTickHandler = new XboxTick();
 Infection* infectionHandler = new Infection();
 FireFight* fireFightHandler = new FireFight();
 HeadHunter* headHunterHandler = new HeadHunter();
-DeviceShop* device_shop = new DeviceShop();
 VariantPlayer* variant_player = new VariantPlayer();
+
+extern int H2GetInstanceId();
+extern XUID xFakeXuid[4];
+std::unordered_map<int, int> object_to_variant;
 
 bool b_H2X = false;
 bool b_GunGame = false;
@@ -35,23 +39,19 @@ bool b_Infection = false;
 bool b_Halo2Final = false;
 bool b_HeadHunter = false;
 
-extern int H2GetInstanceId();
-extern XUID xFakeXuid[4];
-std::unordered_map<int, int> object_to_variant;
+std::unordered_map<wchar_t*, bool&> GametypesMap
+{
+	{ L"h2x", b_H2X },
+	{ L"ogh2", b_XboxTick },
+	{ L"h2f", b_Halo2Final },
+	{ L"gungame", b_GunGame },
+	{ L"zombies", b_Infection },
+	{ L"infection", b_Infection },
+	{ L"wareconomy", b_FireFight },
+	{ L"headhunter", b_HeadHunter },
+	{ L"graverobber", b_HeadHunter }
+};
 
-
-/*constants*/
-const wchar_t* ZOMBIES = L"zombies";
-const wchar_t* INFECTION = L"infection";
-const wchar_t* GUNGAME = L"gungame";
-const wchar_t* H2F = L"h2f";
-const wchar_t* H2X = L"h2x";
-const wchar_t* OGH2 = L"ogh2";
-const wchar_t* GRAVEROBBER = L"graverobber";
-const wchar_t* HEADHUNTER = L"headhunter";
-const wchar_t* WARECONOMY = L"wareconomy";
-const wchar_t* RVB = L"rvb";
-/*         */
 using namespace Blam::Cache::DataTypes;
 
 int GAME_BUILD = 11122;
@@ -1072,64 +1072,25 @@ void __cdecl OnMapLoad(int a1)
 
 		p_set_random_number(a1);
 		return;
-	}
-
-	b_H2X = false;
-	b_GunGame = false;
-	b_XboxTick = false;
-	b_FireFight = false;
-	b_Infection = false;
-	b_HeadHunter = false;
-	b_Halo2Final = false;
+	}		
 
 	wchar_t* variant_name = h2mod->GetLobbyGameVariantName();
 	LOG_TRACE_GAME(L"[h2mod] OnMapLoad engine mode {0}, variant name {1}", h2mod->GetEngineType(), variant_name);
 	BYTE GameState = *(BYTE*)(h2mod->GetBase() + ((h2mod->Server) ? 0x3C40AC : 0x420FC4));
 
+	for (auto gametype_it : GametypesMap)
+		gametype_it.second = false; // reset custom gametypes state
+
 	if (h2mod->GetEngineType() == EngineType::MULTIPLAYER_ENGINE)
 	{
 		addDebugText("GameEngine: Multiplayer");
 
-		// StrStrIW returns pointer to first substring occurrence, NULL othewise
-		if (StrStrIW(variant_name, ZOMBIES) || StrStrIW(variant_name, INFECTION))
+		for (auto gametype_it : GametypesMap)
 		{
-			LOG_TRACE_GAME("[h2mod] Zombies Turned on!");
-			b_Infection = true;
-		}
-
-		if (StrStrIW(variant_name, GUNGAME))
-		{
-			LOG_TRACE_GAME("[h2mod] GunGame Turned on!");
-			b_GunGame = true;
-		}
-
-		if (StrStrIW(variant_name, H2F))
-		{
-			LOG_TRACE_GAME("[h2mod] Halo2Final Turned on!");
-			b_Halo2Final = true;
-		}
-
-		if (StrStrIW(variant_name, H2X))
-		{
-			LOG_TRACE_GAME("[h2mod] Halo 2 Xbox Rebalance Turned on!");
-			b_H2X = true;
-		}
-		else if (StrStrIW(variant_name, OGH2))
-		{
-			LOG_TRACE_GAME("[h2mod] 30 Tick Mod Activated!");
-			b_XboxTick = true;
-		}
-
-		if (StrStrIW(variant_name, GRAVEROBBER) || StrStrIW(variant_name, HEADHUNTER))
-		{
-			LOG_TRACE_GAME("[h2mod] GraveRobber (Headhunter) Turned on!");
-			b_HeadHunter = true;
-		}
-
-		if (StrStrIW(variant_name, WARECONOMY))
-		{
-			LOG_TRACE_GAME("[h2mod] Fire Fight Turned on!");
-			b_FireFight = true;
+			if (StrStrIW(variant_name, gametype_it.first)) {
+				LOG_TRACE_GAME(L"[h2mod] {1} custom gametype turned on!", gametype_it.first);
+				gametype_it.second = true; // enable a gametype if substring is found
+			}
 		}
 
 		get_object_table_memory();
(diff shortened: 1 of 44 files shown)


xlive/H2MOD/Modules/Config/Config.cpp
```diff
@@ -171,7 +171,8 @@ void SaveH2Config() {
 	wchar_t fileConfigPathLog[1124];
 	swprintf(fileConfigPathLog, 1024, L"Saving Config: \"%ws\"", fileConfigPath);
 	addDebugText(fileConfigPathLog);
-	FILE* fileConfig = _wfopen(fileConfigPath, L"wb");
+	FILE* fileConfig; 
+	_wfopen_s(&fileConfig, fileConfigPath, L"wb");
```
num0005: Check the return value and I'm not sure if just switching it like that is ok, _wfopen_s blocks any other program from accessing the file afaik.

xlive/H2MOD/Modules/Config/Config.cpp
```diff
@@ -1855,7 +1856,7 @@ void ReadH2Config() {
 		swprintf(fileConfigPath, 1024, FlagFilePathConfig);
 		swprintf(fileConfigPathLog, 1124, L"Reading Flag Config: \"%ws\"", fileConfigPath);
 		addDebugText(fileConfigPathLog);
-		fileConfig = _wfopen(fileConfigPath, L"rb");
+		_wfopen_s(&fileConfig, fileConfigPath, L"rb");
```
num0005: also here

xlive/H2MOD/Modules/Config/Config.cpp
```diff
@@ -1866,7 +1867,7 @@ void ReadH2Config() {
 			swprintf(fileConfigPath, 1024, H2ConfigFilenames[H2IsDediServer], checkFilePath, readInstanceIdFile);
 			swprintf(fileConfigPathLog, 1124, L"Reading Config: \"%ws\"", fileConfigPath);
 			addDebugText(fileConfigPathLog);
-			fileConfig = _wfopen(fileConfigPath, L"rb");
+			_wfopen_s(&fileConfig, fileConfigPath, L"rb");
```
num0005: and here

xlive/XLive/Networking/XSocket.cpp
```diff
@@ -19,6 +21,10 @@ int WINAPI XWSACleanup()  	// XWSACleanup
 int WINAPI XSocketClose(SOCKET s)
 {
 	LOG_TRACE_XLIVE("XSocketClose: socket: {}", s);
+
+	if (userManager.sockmap[s])
```
num0005: that will insert an element, use `count` or `find`

xlive/XLive/Networking/xlive_network.cpp
```diff
@@ -266,21 +265,12 @@ int WINAPI XSocketRecvFrom(SOCKET s, char *buf, int len, int flags, sockaddr *fr
...
-				if (user)
-				{
-					if (user->xnaddr.ina.s_addr != iplong)
-					{
-						user->xnaddr.ina.s_addr = iplong;
-					}
-				}
-				else // only create the user if they're not already in the system.
-					userManager.CreateUser((XNADDR*)(buf + 8), TRUE);
+				XNADDR* userxn = reinterpret_cast<XNADDR*>(buf + 8);
```
num0005: not your fault but would it be possible to use a struct for buf? It seems we know what's at every offset.

xlive/xliveless.cpp
```diff
@@ -187,7 +187,7 @@ BOOL SetDlcBasepath( int num )
 
 					// check file exists
 					FILE *fp_dlc;
-					fp_dlc = _wfopen( wnum.c_str(), L"rb" );
+					_wfopen_s(&fp_dlc, wnum.c_str(), L"rb" );
```
num0005: this file also has the issue with not checking the return value.

xlive/XLive/UserManagement/CUser.cpp
```diff
@@ -51,13 +49,29 @@ void CUserManagement::CreateUser(const XNADDR* pxna, BOOL user)
...
-	nUser->secure.s_addr = secure;
+	CUser* nUser = cusers[secure];
+	// check if the user is already in the system
+	if (nUser == nullptr)
+	{
+		// allocate new mem for the new peer
+		LOG_TRACE_NETWORK("CUserManagement::CreateUser() new secure address {:x}", pxna->inaOnline.s_addr);
+
+		nUser = new CUser;
+		memset(&nUser->xnaddr, 0x00, sizeof(XNADDR));
```
num0005: why are you clearing the addr if you are going to memcpy over it?

xlive/XLive/UserManagement/CUser.cpp
```diff
@@ -98,51 +112,41 @@ void CUserManagement::CreateUser(const XNADDR* pxna, BOOL user)
...
 	if (to_remove_user != nullptr)
 	{
-		std::pair<ULONG, short> smap, smap1;
-		if (to_remove_user->xnaddr.ina.s_addr == H2Config_ip_wan)
-		{
-			smap = std::make_pair(H2Config_ip_lan, to_remove_user->xnaddr.wPortOnline);
-			smap1 = std::make_pair(H2Config_ip_lan, ntohs(htons(to_remove_user->xnaddr.wPortOnline) + 1));
-		}
-		else
+		if (xnmap[ina.s_addr])
```
num0005: this also could insert unneeded elements into the array, use `count` or `find`, etc

xlive/XLive/UserManagement/CUser.cpp
```diff
@@ -98,51 +112,41 @@ void CUserManagement::CreateUser(const XNADDR* pxna, BOOL user)
...
+
+		if (cusers[ina.s_addr])
+			cusers.erase(ina.s_addr);
+
+		for (auto it : secure_map)
 		{
-			smap = std::make_pair(to_remove_user->xnaddr.ina.s_addr, to_remove_user->xnaddr.wPortOnline);
-			smap1 = std::make_pair(to_remove_user->xnaddr.ina.s_addr, ntohs(htons(to_remove_user->xnaddr.wPortOnline) + 1));
+			if (it.second == ina.s_addr) {
+				secure_map.erase(it.first);
```
num0005, line 148: not sure if it's safe to remove elements while iterating? Could crash or skip an element... Might be better to use std::remove_if

xlive/H2MOD/Modules/Config/Config.cpp
nukeulater: The files get closed after read/write, blocking shouldn't be a problem.

xlive/XLive/UserManagement/CUser.cpp
nukeulater: it exits the loop after it finds the element, shouldn't cause any issues

xlive/XLive/Networking/xlive_network.cpp
```diff
@@ -262,25 +261,17 @@ int WINAPI XSocketRecvFrom(SOCKET s, char *buf, int len, int flags, sockaddr *fr
 		u_long iplong = (((struct sockaddr_in*)from)->sin_addr.s_addr);
 		std::pair <ULONG, SHORT> hostpair = std::make_pair(iplong, port);
 
+		SecurePacket* secure_pck = reinterpret_cast<SecurePacket*>(buf);
 		if (iplong != H2Config_master_ip)
 		{
-			if (*(ULONG*)buf == annoyance_factor && ret == 12 + sizeof(XNADDR))
+			if (ret == sizeof(SecurePacket)
```
num0005: would be nicer to call it something other than `ret`, maybe `packet_size`. Could also merge the two if statements.

xlive/XLive/UserManagement/CUser.cpp
```diff
@@ -204,13 +209,12 @@ void CUserManagement::ConfigureLocalUser(XNADDR* pxna, ULONGLONG xuid, char* use
 	memset(&local_user, NULL, sizeof(CUser));
 	memcpy(&local_user.xnaddr, pxna, sizeof(XNADDR));
 	local_user.secure.s_addr = pxna->inaOnline.s_addr;
-	this->secure_packet = new char[12 + sizeof(XNADDR)];
-	memset(secure_packet, NULL, 12 + sizeof(XNADDR));
+	memset(&securePacket, NULL, sizeof(SecurePacket));
```
num0005: using `0` instead of `NULL` or the `ZeroMemory` macro would make it more clear. `NULL` could make it seem like you are dealing with an array of pointers.

xlive/xliveless.cpp
```diff
@@ -972,8 +972,7 @@ DWORD WINAPI XStorageUploadFromMemory( DWORD dwUserIndex, const WCHAR *wszServer
 
 
 	FILE *fp;
-
-	fp = _wfopen( wszServerPath, L"wb" );
+	_wfopen_s(&fp, wszServerPath, L"wb" );
```
num0005: still need to check return value here, per MSDN

xlive/XLive/UserManagement/CUser.cpp
num0005: ok yeah that's fine then.

num0005:
You might want to consider using an enum for GametypesMap, so other code doesn't need to know the exact string it checks against.

xlive/XLive/Networking/xlive_network.cpp
nukeulater: the if statements can't be merged

nukeulater:
Not sure I understand.

xlive/XLive/Networking/xlive_network.cpp
num0005: I missed the rest of the code that handles packets that aren't SecurePacket

xlive/XLive/UserManagement/CUser.cpp
```diff
@@ -12,15 +10,17 @@
 extern XUID xFakeXuid[4];
 extern CHAR g_szUserName[4][16];
 
+SecurePacket securePacket;
 CUserManagement userManager;
-const DWORD annoyance_factor = 0x11223344;
+const DWORD annoyance_factor = 0x8E0A40F1;
```
shayaantx: uhhh, please remove

xlive/XLive/UserManagement/CUser.cpp
shayaantx: is it fine though? how many different threads might access this map at the same time? if we can guarantee single threaded access to the map then it seems okay, but in reality it should be thread safe

xlive/H2MOD/Modules/Networking/CustomPackets/CustomPackets.cpp
```diff
@@ -162,13 +164,13 @@ void __stdcall message_gateway_hook_2(void *thisx, int a2, int message_type, int
...
+	case request_map_filename:
 	{
 		s_request_map_filename* pak = (s_request_map_filename*)packet;
-		LOG_TRACE_NETWORK("[H2MOD-CustomPackets] received on hook2 request-map-filename from XUID: {:#x}", pak->user_identifier);
+		LOG_TRACE_NETWORK("[H2MOD-CustomPackets] received on message_gateway_hook_2 request-map-filename from XUID: {:#x}", pak->user_identifier);
 		return;
 	}
-	case e_network_message_types::map_file_name:
+	case map_file_name:
 		return;
```
shayaantx, line 174: so for map_file_name we don't want to call p_network_message_gateway_2 but for everything other type we do? 

there should be a single line doc explaining this and why

xlive/H2MOD.cpp
```diff
@@ -293,7 +292,7 @@ signed int __cdecl object_new_hook(void *pObject)
 	object_to_variant[result] = variant_index;
 
 	//wchar_t DebugText[255] = { 0 };
-	//ZeroMemory(DebugText, sizeof(DebugText));
```
shayaantx, line 296: please revert back to ZeroMemory unless you have a real reason to secure this data

xlive/H2MOD.cpp
```diff
@@ -881,7 +877,7 @@ void H2MOD::CustomSoundPlay(const wchar_t* soundName, int delay)
 		if (delay > 0)
 			std::this_thread::sleep_until(timePoint);
 
-		PlaySound(soundName, NULL, SND_FILENAME);
+		PlaySound(soundName, NULL, SND_FILENAME | SND_NODEFAULT);
```
shayaantx, line 880: when is soundName null/empty? why even go through any of this code if a null soundName was passed in? check add a null check and return

xlive/H2MOD.cpp
```diff
@@ -36,23 +39,19 @@ bool b_Infection = false;
 bool b_Halo2Final = false;
 bool b_HeadHunter = false;
 
-extern int H2GetInstanceId();
-extern XUID xFakeXuid[4];
-std::unordered_map<int, int> object_to_variant;
-
+std::unordered_map<wchar_t*, bool&> GametypesMap
```
shayaantx, line 42: In reference to https://github.com/PermaNulled/cartographer/pull/242#issuecomment-514374292 hes probably right, an enum would be cleaner, however...

What you want to stride for is not having a custom map, you should be thinking, if 2moro someone wanted to add a new game type in code how many different spots would they need to update.

It'd be very nice if this map was driven by all the classes that implement GameClientServerHandler (i.e., register each sub class of GameClientServerHandler in some map). Then embed the string constant and whether its enabled in those classes as well. That way adding a new game type just involves implementing GameClientServerHandler. This would be the polymorphic way of handling this

You can even improve a lot of the event code (like player spawn and player death), for example

Right now there is something like below

```
if (b_HeadHunter)
	{
		DatumIndex dead_player = DatumIndex(unit_datum_index);
		headHunterHandler->playerDeath->SetDeadPlayer(dead_player); // set this so we can spawn a skull on their position.
		headHunterHandler->playerDeath->execute();
	}
	if (b_GunGame) {
		gunGame->playerDeath->setUnitDatumIndex(unit_datum_index);
		gunGame->playerDeath->execute();
	}
```

but it really should be

```	
for (auto handler : game_handlers) {
		if (handler.enabled()) {
			handler->playerDeath.SetDeadPlayer(dead_player);
			handler->playerDeath->execute();
		}
	}
```

You can get rid of a lot of duplicated code like this

PR #259, opened 2020-04-14
Implement universal fire rate calculation

ShellyHerself:
This was sparked from a discussion involving me and General_101. I've dealt with the opposite problem before in CE as that game's tickrate IS locked. So instead of firerates being higher, they are correct to Xbox, which can make them slower than intended for user mods.

What this does is it implements the universal calculation of the effective firing rates on Xbox and applies those fire rates to the tags.

I did have to add a small hack to make my recovery rates match what you guys had put in. And I am convinced that while it alligns with your values, it may be wrong compared to Xbox. Please test and if the guns are a little too slow, do what the comment tells you to do.

Finally, I don't have visual studio and am not even on Windows right now. So, you may have small missing details like semicolons and stuff. I've set the pullrequest to be editable by maintainers.

Here if some extra reading material for if the changes don't really make sense to you https://gist.github.com/gbMichelle/927384281567e43e413fda761fb846fc

I can answer any questions you may have. Have a nice day!

### DIFF
--- a/xlive/H2MOD/Variants/H2X/H2X.cpp
+++ b/xlive/H2MOD/Variants/H2X/H2X.cpp
@@ -70,7 +70,7 @@ void H2X::Initialize(bool enable)
 
 				// Go through all barrel blocks and apply the fix.
 				for (int i = 0; i < barrel_data_block->block_count; i++) {
-					barrel_block = barrel_block + barrel_data_block_size * i;
+					auto barrel_block = barrel_block + barrel_data_block_size * i;
 
 					float* firerate_lower = (float*)(barrel_block + 4);
 					float* firerate_upper = (float*)(barrel_block + 8);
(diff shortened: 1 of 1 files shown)


xlive/H2MOD/Variants/H2X/H2X.cpp
```diff
@@ -19,11 +27,53 @@ std::vector<H2X::h2x_mod_info> weapons =
...
+	if (h2v_firerate < 0.0001) {
+		// Do not divide by 0.
+		return h2v_firerate;
+	} else if (h2v_firerate > 30.0) {
+		/* Firing rates above 30 are not possible at 30fps and below. */
+		return 30.0;
+	}
+	/* - 0.1 to avoid a really pesky floating point rounding error
+	   This error and rounding hack only affects higher values above 10.*/
+	return 30.0 / ceil(1 / h2v_firerate * 30.0 - 0.1);
```
nukeulater: any reason to declare these values as doubles instead of floats (also that 1 declared as int)? (it could be the root problem with rounding)

ShellyHerself: I have thought about it. Current testing has showed it is only a problem in final builds. Not debug builds, showing that there is an issue with how the math is being optimized by the compiler.

I can make a version of the code with it being converted to a double first.

ShellyHerself:
I've updated the code to use doubles for most of the math that might need it. I've lowered the number for the rounding error prevention. If a release build has a battle rifle that fires it's individual shots every three frames instead of every two then I'm putting back the aggressive rounding fix while keeping the doubles.

PR #285, opened 2020-09-23
Fixed badly generated XML

Kantanomo:
Visual Studio did not generate the XML properly when some class files were made, this was causing the filters to be invalidated. which would force rebuild the file.


Kantanomo:
@iSupersniper This should fix the issue.

Kantanomo:
0

iSupersniper:
Dunno what this is, Good Bye. :)

PR #297, opened 2021-03-29
Before Revert and all nukes changes, backported 6.2.1 code into it.

Kantanomo:


nukeulater:
Details:

Fixed built in frame limiters and stuttering by setting the global timer resolution.
More info here https://randomascii.wordpress.com/2020/10/04/windows-timer-resolution-the-great-rule-change/
Applies to Windows 10 v2004 and above (maybe bellow as well, if other applications set the global resolution timer)

Fixed serverlist buffer size calculation and a memory leak when handling strings.

Fixed LIVE serverlist from dropping UI labels when hovering over player populated servers.

XNet code improvements and implemented some other XLive APIs better.

Fixed performance when using high polling rate mice, causing the game to drop framerate when moving around when the cursor is visible, by limiting how many times the cursor indicator can be set to 125hz maximum on each WM_SETCURSOR window procedure message.

Improved performance of keyboard state polling by using just one call to GetKeyboardState instead of calling GetAsyncKeyState 256 times.

Fixed a game bug when tickrate is increased from 30.

improved the logging code, now logs can always be created at each point, the code is not hardcoded by H2Configs anymore, also re-enabled OnScreenDebug to log everytime even when logs are disabled by default, and added always enabled Critical Network log to log any critical network issues.

Removed Infection hardcoded weapons, they can be set in the playlist.

And misc code cleanup

PR #299, opened 2021-03-30
fixed scenario/engine type

nukeulater:
also replaced h2mod->Server with Memory::IsDedicatedServer

### DIFF
--- a/xlive/H2MOD/Tags/MetaLoader/tag_loader.cpp
+++ b/xlive/H2MOD/Tags/MetaLoader/tag_loader.cpp
@@ -107,7 +107,7 @@ namespace tag_loader
 		fin->seekg(0x0);
 		fin->read(map_header, 0x800);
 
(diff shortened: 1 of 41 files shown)


xlive/H2MOD/Tags/MetaLoader/tag_loader.cpp
```diff
@@ -1349,7 +1349,7 @@ bool _cdecl LoadTagsandMapBases(int a)
 	///tag_injector testing
 	//Just for testing purpose,dont cluter here	
 	///Actual injection process after map load
-	if (tags::get_cache_header()->type != e_engine_type::SinglePlayerShared)
+	if (tags::get_cache_header()->is_single_player_shared())
```
Himanshu-01: should be
if (!tags::get_cache_header()->is_single_player_shared())

nukeulater: fixed

PR #315, opened 2021-05-30
fix frame limiter setting not applying after reading config

nukeulater:
warning fix, also hopefully fixed some file encoding error of AdvancedSetings.cpp due to spanish symbols


NukeULater:
> warning fix, also hopefully fixed some file encoding error of AdvancedSetings.cpp due to spanish symbols

PR #338, opened 2021-09-29
Add Extra Graphics Options

Berthalamew:
Also disabled 60fps cinematic patch since it broke camera animations


xlive/H2MOD/Modules/TagFixes/TagFixes.cpp
```diff
@@ -134,6 +136,38 @@ namespace TagFixes
...
+			{
+				auto shader = tags::get_tag<blam_tag::tag_group_type::shader, shader_definition>(shader_item.first);
+				shader->shaderLODBias = shader_definition::_Never;
+			}
+		}
+
+
+		//Hacky thingy that does stuff for setting flags to true or false
+		//Should Prob Remove when a better solution is found
+		template<class T> inline T operator~ (T a) { return (T)~(int)a; }
```
nukeulater, line 153: no idea why do you even need these?

xlive/H2MOD/Modules/RenderHooks/RenderHooks.cpp
```diff
@@ -93,10 +97,48 @@ namespace RenderHooks
 					return 4096;
 			}
 		}
+		if (e == e_layer_type::bloom)
```
nukeulater, line 100: these should be a switch of e_layer_type (s)

PR #344, opened 2021-10-15
Event Handler fix

Kantanomo:
I'm not the brightest crayon, events not matching the correct type were being called. fixed by implementing an enum into the calls


nukeulater:
i added another fix, i'll close this

PR #411, opened 2023-02-22
Restrict uncapped modes.

Kantanomo:


xlive/H2MOD/Modules/Shell/Config.cpp
```diff
@@ -724,12 +724,12 @@ void ReadH2Config() {
...
+					/*case 1:
 						H2Config_experimental_fps = _rendering_mode_old;
 						break;
 					case 2:
 						H2Config_experimental_fps = _rendering_mode_new;
-						break;
+						break;*/
 					case 3:
 						H2Config_experimental_fps = _rendering_mode_original_game_frame_limit;
 				}
```
Berthalamew: Can we clean this up so we don't have dead code lying around, so we just have options 0 and 1 for none and _rendering_mode_original_game_frame_limit

Berthalamew (changes requested):

xlive/H2MOD/Modules/Shell/Config.cpp
Kantanomo: Want the functionality to be easily re-addable?

xlive/H2MOD/Modules/Shell/Config.cpp
```diff
@@ -721,17 +721,14 @@ void ReadH2Config() {
 				switch(std::stoi(ini.GetValue(H2ConfigVersionSection.c_str(), "experimental_rendering", "0")))
 				{
 					default:
+						//Incase any of the old rendering modes were used for a higher fps, set it back to 60.
+						H2Config_fps_limit = 60;
 					case 0:
 						H2Config_experimental_fps = _rendering_mode_none;
 						break;
```
Berthalamew, line 728: the statement inside of default should be in case 0: and both should execute it

Berthalamew (changes requested):

xlive/H2MOD/Modules/Shell/Config.cpp
twist84: default falls through into case 0

Berthalamew: was under the impression that we were limiting it to 60 if we were using _rendering_mode_none, so we wouldn't have complaints about slowdown.
Clarified with kant and it's only for resetting the fps limit for people who had the other modes set.

Berthalamew (approved):

PR #416, opened 2023-03-23
Reimplement object_new

Berthalamew:
rebased PR based on #414 

- reimplements object_new function and most of the subcalls it makes
- needed to fix issues with object creation in coop
- extensive testing required before we merge this


Berthalamew:
Closed since issue was identified without object_new rewrite
structs and other useful things from this PR will be merged later on

PR #421, opened 2023-03-29
Fix zombies vehicle spawning issue

Berthalamew:
- change classification of netgame items to a powerup if the classification is already a vehicle or if it's using a vehicle collection
- remove some unused and duplicated code
- fixed issue with callbacks to GameLoop on dedicated servers


xlive/H2MOD/Modules/MainLoopPatches/RunLoop/RunLoop.cpp
Kantanomo, line 1: There is no reason to remove these, kill volume stuff will be used in the future.

Berthalamew: I'll move this into another module and clean up the functions

xlive/H2MOD/Modules/MainLoopPatches/RunLoop/RunLoop.cpp
```diff
@@ -235,6 +235,15 @@ void __cdecl main_game_loop_hook() {
 	EventHandler::GameLoopEventExecute(EventExecutionType::execute_after);
 }
 
+typedef void(_cdecl* main_loop_body_t)();
+main_loop_body_t p_main_loop_body_dedi;
+void __cdecl main_loop_body_dedi()
```
nukeulater: I think it's possible to use `main_game_loop_hook` instead of creating another function. `shouldQuitMainLoop` and `dedicated_server_should_quit_hook` can then be safely removed.

Kantanomo: Yeah should work to just combine the two, and add a check for map downloader and local user stats to only run on client.

xlive/H2MOD/Variants/Infection/Infection.cpp
```diff
@@ -259,10 +225,51 @@ void Infection::onGameTick()
...
+			{
+				itmc->item_permutations[i]->item.TagGroup = blam_tag::tag_group_type::equipment;
+				itmc->item_permutations[i]->item.TagIndex = shotgun_ammo_equip_datum;
+			}
+		}
+	}
+
+	//Replace vehicles with shotgun ammo
+	s_scenario_group_definition* scenario = get_global_scenario();
+	for (DWORD i = 0; i < scenario->netgame_equipment.size; i++)
```
Kantanomo, line 252: I fail to understand what this change accomplishes it ends up doing the same thing.

Berthalamew: If you're referring to getting the scnr, it does the same thing but instead of creating a for loop and searching for the scnr tag we can just grab the global one

Kantanomo: no the entire thing.

Berthalamew: If the collection ref is null and the classification is a vehicle, vehicles can still spawn even when that ref is null since it'll pull a default from the gametype

Kantanomo: That was intentionally not fixed for vairants that explicitly wanted vehicles.

Berthalamew: Changed

xlive/H2MOD/Modules/MainLoopPatches/RunLoop/RunLoop.cpp
Berthalamew: Combined the two main_loop_body functions and removed shouldQuitMainLoop as well as dedicated_server_should_quit_hook

Kantanomo (approved):
Changes requested were made, looks good to me just unless Nuke sees something wrong.

except... 
`const kill_volume_enable_or_disable_t func = (enable_volumes ? scenario_kill_trigger_volume_enable : scenario_kill_trigger_volume_disable);`

could just be an if statement and no need for the global typedef.

nukeulater (approved):

PR #424, opened 2023-04-07
Migration from ini config to a Json config

Kantanomo:
Created a class built on top of RapidJson to allow easy access and storage of values inside of a Json File. Class can be used anywhere in the project and is easy to implement anywhere else inside the project.

Main benefits of this change include

- Cleaner settings file with organization of settings.
- Ease of adding new settings to the project
- Code for adding and accessing settings is simplified and has good readability
- Adding support for custom types is very easy

One downside of the update means that we cannot place comments into the config file to give people a description of the settings file.


xlive/H2MOD/Modules/Shell/JsonConfig/JsonConfig.cpp
```diff
@@ -0,0 +1,253 @@
...
+
+        Value& v = (*current_object)[key];
+        if (!current_object->HasMember(key)) {
+            if (defaultValues_.find(key) != defaultValues_.end())
+                v = defaultValues_[key];
+            else
+                return defaultValue;
+        }
+
+    	if constexpr (std::is_same_v<T, bool>) {
```
nukeulater: I think you can remove all of these `constexpr` specializations by simply using `return v.Get<T>();`. And some validation wouldn't hurt because just the type alone doesn't guarantee that the value/object in the json document is of the same type.

xlive/H2MOD/Modules/Shell/JsonConfig/JsonConfig.cpp
```diff
@@ -0,0 +1,253 @@
+#include "stdafx.h"
+#include <fstream>
```
nukeulater: move these std libraries to `stdafx.h` if they're not already included there

xlive/H2MOD/Modules/Shell/Config.cpp
```diff
@@ -51,7 +54,7 @@ H2Config_Deadzone_Type H2Config_Controller_Deadzone = H2Config_Deadzone_Type::Ax
 float H2Config_Deadzone_A_X = 26.0f;
 float H2Config_Deadzone_A_Y = 26.0f;
 float H2Config_Deadzone_Radial = 1.0f;
-float H2Config_crosshair_offset = NAN;
+float H2Config_crosshair_offset = 0.138f;
```
nukeulater: default h2v is 0.152 or .156 iirc

xlive/H2MOD/Modules/Shell/JsonConfig/JsonConfig.cpp
Kantanomo: Get<T> had compilation issues, I'll revisit it.

xlive/H2MOD/Modules/Shell/Config.cpp
Berthalamew: the default is 0.165

xlive/H2MOD/Modules/Shell/JsonConfig/JsonConfig.cpp
Kantanomo: Merged all of the base RapidJson value types into a single Get<T> call while leaving the expressions for the custom types.

Added Json Value validation that will check if there was a default parameter passed to it or return from the defaultValue store if it doesn't exist. If neither exist it will just return the default value of the specific type.

xlive/H2MOD/Modules/Shell/Config.cpp
Kantanomo: fixed.

PR #429, opened 2023-04-25
Animation research 

Berthalamew:
- defined the animation graph definition as well as the member classes of it and some of the member functions
- update definitions for c_animation_channel and c_interpolator_control
- fixes compilation issues with constant tagblocks and string ids

- this is beneficial since it allows us to read and access animation tags incase someone wants to use them later on in project


xlive/Blam/Common/Common.h
```diff
@@ -42,7 +42,7 @@ static_assert (sizeof(STRUCT) == _SIZE, "Invalid size for struct (" #STRUCT")")
 static_assert (offsetof(STRUCT, FIELD) == OFFSET, #STRUCT " layout is invalid")
 
 #define TAG_BLOCK_SIZE_ASSERT(tagblock,size)\
-static_assert (sizeof(tagblock) == (size),"Invalid Size for TagBlock <" #tagblock ">")
+static_assert (sizeof(tagblock) == size,"Invalid Size for TagBlock <" #tagblock ">")
```
nukeulater: why?

Berthalamew: Was looking at the assert while working on the definitions and the brackets seemed unnecessary.
I can add them back if you want.

xlive/Blam/Engine/animations/animation_definitions.cpp
```diff
@@ -0,0 +1,303 @@
...
+}
+
+e_animation_type c_model_animation::get_animation_type() const
+{
+	return this->type;
+}
+
+float c_model_animation::get_authored_duration() const
+{
+	return this->frame_count * 0.0333333333f;
```
nukeulater: `frame_count / 30.0f` or `frame_count * (1.0f / 30.0f)` would be more preferable

xlive/Blam/Common/Common.h
nukeulater: macro pitfalls: https://gcc.gnu.org/onlinedocs/gcc-3.0.1/cpp_3.html#SEC26

xlive/Blam/Engine/animations/animation_definitions.cpp
Berthalamew: changed to this->frame_count * (1.0f / 30.0f)

xlive/Blam/Common/Common.h
Berthalamew: reverted back to before

PR #430, opened 2023-05-01
Revamp HUD code

Berthalamew:
- cleanup and modify most of the logic from the HudElements that was moved out to their own modules
- add structs s_new_hud_globals and s_new_hud_globals_player_info (not sure about these names)
- fix crosshair scaling code so that it doesn't down sample the bitmaps for the scaling
- add code to allow us to scale UI based on a resolution modifier (for future use)

- beneficial since the previous code had issues with resolution scaling and it uses the proper folder structures for the in-game logic
- code for resolution scaling is planned to be used in the future for a hud redraw that sako has partially started and offered to complete


xlive/Blam/Engine/interface/first_person_camera.cpp
```diff
@@ -0,0 +1,84 @@
...
+	float calculated_radians_FOV;
+	if (currentVariantSettings.forcedFOV == 0)
+	{
+		calculated_radians_FOV = fov * M_PI / 180.0f;
+	}
+	else
+	{
+		calculated_radians_FOV = currentVariantSettings.forcedFOV * M_PI / 180.0f;
+	}
+	WriteValue(Memory::GetAddress(0x413780, 0x3B5300), calculated_radians_FOV);
```
nukeulater: use `WriteValue` only when writing to a write protected memory segment/page

xlive/Blam/Engine/interface/first_person_camera.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+#include "H2MOD/Modules/CustomVariantSettings/CustomVariantSettings.h"
+#include "H2MOD/Modules/Shell/Config.h"
+#include "H2MOD/Tags/TagInterface.h"
+#include "Util/Hooks/Hook.h"
+
+#define _USE_MATH_DEFINES
+#include <math.h>
+
+static bool b_showHUD = true;
```
nukeulater: we should remove all hungarian style bool prefixes

xlive/Blam/Engine/interface/motion_sensor.cpp
```diff
@@ -0,0 +1,25 @@
...
+	if (Memory::IsDedicatedServer()) { return; }
+
+	radar_patch();
+	motion_sensor_fix_size();
+}
+
+void motion_sensor_fix_size()
+{
+	float og_res = *Memory::GetAddress<float*>(0x3E6A9C);
+	WriteValue<float_t>(Memory::GetAddress(0x3E6A9C), og_res * SIZE_MULTIPLIER);
```
nukeulater: checked the code and `hud_scale` takes care of this, why is this needed?

Berthalamew: This increases the resolution of the pulse so that it it will be the correct size on the screen.
If the resolution is not modified the pulse will be smaller than the radar bitmap when the proper size is used

ex. I've used a radar bitmap that's 4x bigger than before when modifying the hud_scale, the blip is still scaled 4x lower than normal
image

scaling the value properly fixes this
image

xlive/Blam/Engine/interface/first_person_camera.cpp
Berthalamew: Changed the following to 
*Memory::GetAddress<float*>(0x413780, 0x3B5300) = calculated_radians_FOV;

xlive/Blam/Engine/interface/first_person_camera.cpp
Berthalamew: that bool wasn't supposed to be there but changed the name of the actual one

xlive/Blam/Engine/interface/hud.cpp
```diff
@@ -168,7 +130,11 @@ void hud_apply_pre_winmain_patches()
 
 	p_update_hud_elements_display_settings = Memory::GetAddress<update_hud_elements_display_settings_t>(0x264A18, 0x0);
 
+	// Replace all calls to update_hud_elements_display_settings with our hook
 	PatchCall(Memory::GetAddress(0x25E1FC, 0x0), update_hud_elements_display_settings_hook);
 	PatchCall(Memory::GetAddress(0x264058, 0x0), update_hud_elements_display_settings_hook);
 	PatchCall(Memory::GetAddress(0x26406F, 0x0), update_hud_elements_display_settings_hook);
+
+	// Replace the crosshair and text scale global with our own just for the crosshair
+	WriteValue(Memory::GetAddress(0x222F9F + 4), &crosshair_size);
```
nukeulater: `WritePointer()` here

xlive/H2MOD/Modules/Shell/Startup/WinMainH2.cpp
```diff
@@ -94,6 +92,9 @@ int WINAPI WinMain_Halo2(HINSTANCE hInstance, HINSTANCE hPrevInstance, LPSTR lpC
 	// mess around with xlive (not calling XLiveInitialize etc)
 	WriteValue<BYTE>(Memory::GetAddress(0x4FAD98), 1);
 
+	// Apply patches for the hud that need to be applied before WinMain is called
+	hud_apply_pre_winmain_patches();
```
nukeulater: move to tweaks, there patches are applied before startup

xlive/Blam/Engine/interface/hud.cpp
```diff
@@ -168,7 +130,11 @@ void hud_apply_pre_winmain_patches()
 
 	p_update_hud_elements_display_settings = Memory::GetAddress<update_hud_elements_display_settings_t>(0x264A18, 0x0);
 
+	// Replace all calls to update_hud_elements_display_settings with our hook
 	PatchCall(Memory::GetAddress(0x25E1FC, 0x0), update_hud_elements_display_settings_hook);
```
nukeulater: extra 0x0 param is redundant here

xlive/Blam/Engine/interface/hud.cpp
```diff
@@ -0,0 +1,140 @@
...
+	float g_ui_scale = *Memory::GetAddress<float*>(0xA3E424);
+	if (a2 && a2 != a1) // if a2 == a1 then this is a chapter name which scaling seems to break, so skip those ones
+	{
+		short width = a2->right - a2->left;
+		a2->right = a2->left + (short)((float)width * g_ui_scale); // scale width by UI_Scale, stops text elements from being cut-off when UI_Scale > 1
+	}
+	return p_render_text_wrapper(a1, a2, a3, a4, a5, a6, a7, a8);
+}
+
+void set_hud_size(const float size)
```
nukeulater: `const` parameters should mainly be used for references or pointers to variables that shouldn't be modified

nukeulater: otherwise is just extra syntax sugar

Berthalamew: changed

xlive/Blam/Engine/interface/hud.cpp
Berthalamew: removed

xlive/H2MOD/Modules/Shell/Startup/WinMainH2.cpp
Berthalamew: moved it to tweaks

xlive/Blam/Engine/interface/hud.cpp
Berthalamew: changed to WritePointer

PR #431, opened 2023-05-16
cleanup high res patch

Berthalamew:
- moved patches to their respective modules
- remove old high res fix code that wont be used anymore
- clean the code to use proper structs, names, and pre-existing functions to get the various hud scales

### DIFF
--- a/xlive/Blam/Engine/interface/hud.cpp
+++ b/xlive/Blam/Engine/interface/hud.cpp
@@ -8,11 +8,11 @@
 #include "H2MOD/Utils/Utils.h"
 #include "Util/Hooks/Hook.h"
 
-static float original_crosshair_text_scale;
-static float original_hud_scale;
+static float original_primary_hud_scale;
+static float original_secondary_hud_scale;
 
 // Seperated value for the scale of the crosshair
-static float crosshair_size = 1.0f;
+static float crosshair_scale = 1.0f;
 
 // This controls the scale for hud when we implement the redrawn hud
 // ex. a value of 4.0 would allow us to use hud bitmaps that are 4x larger (resolution wise) than the old ones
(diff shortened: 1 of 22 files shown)


xlive/Blam/Engine/interface/hud.cpp
```diff
@@ -90,6 +43,11 @@ void set_crosshair_and_text_size(float size)
 	*Memory::GetAddress<float*>(0x464028, 0x0) = original_crosshair_text_scale * size * (1 / k_hud_upscale_size);
 }
 
+float* get_crosshair_and_text_size()
```
nukeulater: `get_crosshair_and_text_size` ? pretty sure this is hud scale.

Berthalamew: it scales the text, crosshair, and the scoreboard.
the actual hud scale is at 0x46402C

nukeulater: Well yeah, it scales some elements of the hud, I'm not necessarily talking about what it does exactly but what it actually represents, in this case 'hud_scale'.
Can't check rn but the other offset is probably the `hud_size`, which is practically `hud_scale` (by res) multiplied by `size` = `hud_size`. hud size scaled by res to be more specific.

Berthalamew: I used scale and size interchangeably when naming the globals and functions
It would probably be better to have the crosshair_and_text_size named secondary_hud_scale and the regular hud_scale called primary_hud_scale.

Unless you've got a better name since both scales control different parts of the hud.

nukeulater: That would be better.

Berthalamew: changed the globals and the function names

PR #432, opened 2023-05-25
cleanup hud definitions and related patches

Berthalamew:
- cleaned up hud_globals definition
- cleaned up wgit definition
- move screen modules to proper location
- cleanup and fix issues with xbox live progress menu (text and model were not in the correct location)
- remove pregame lobby patches (these were never used)
- move ESRB patch to its own module


nukeulater:
lgtm, @Himanshu-01 ?

xlive/Blam/Engine/interface/screens/screen_xbox_live_task_progress_dialog.cpp
```diff
@@ -1,78 +1,56 @@
...
+	auto p_c_screen_xbox_live_task_progress_dialog2 = Memory::GetAddress<c_screen_xbox_live_task_progress_dialog_t>(0x20C776);
 
-	p_c_screen_xbox_live_task_progress_dialog2(-1, 0, callback, 0, 0);
+	p_c_screen_xbox_live_task_progress_dialog2(-1, 0, update_function, 0, 0);
 }
 
-void c_xbox_live_task_progress_menu::ApplyPatches()
+void c_xbox_live_task_progress_menu::apply_patches()
 {
-	if (h2mod->GetEngineType() != _main_menu || Memory::IsDedicatedServer()) return;
```
Himanshu-01: This line needs to be present there as you are not checking it inside `screens_apply_patches_on_map_load();` either

xlive/Blam/Engine/interface/screens/screens_patches.cpp
```diff
@@ -1,10 +1,15 @@
...
 #include "screen_xbox_live_task_progress_dialog.h"
 
 void screens_apply_patches_on_map_load()
 {
 	c_xbox_live_task_progress_menu::apply_patches();
+}
 
+void screens_apply_hooks()
+{
+	c_screen_esrb_warning::apply_patches();
```
Himanshu-01, line 14: None of the codes inside `c_screen_esrb_warning::apply_patches();`  are related to `hooking` so maybe rename this function as `screens_apply_patches_on_startup` ?

xlive/Blam/Engine/interface/screens/screen_xbox_live_task_progress_dialog.cpp
Berthalamew: The mainmenu check shouldn't matter since when it tries to find the tag datum on a map other than the mainmenu it'll return -1 for the task_progress_dialog_datum_index and return

nukeulater: Yeah, but what is the point of going through all the tag paths just to find out the tag is not there? Literally a waste of CPU cycles and implicitly of electricity. Should add them back.

Berthalamew: moved the function so it doesn't execute unless its not a dedi and its the mainmenu engine

xlive/Blam/Engine/interface/screens/screens_patches.cpp
Berthalamew: changed the function name to screens_apply_patches

PR #434, opened 2023-05-29
don't override fov when it's set to 0

Berthalamew:
- players shouldn't be able to set it to 0 within the menus but add check for people who set it manually.
- remove text in config for the hiresfix
- set default fov to 78 instead of 70 in the config (we already have the default option in imgui set to 78)


nukeulater:
> set default fov to 78 instead of 70 in the config (we already have the default option in imgui set to 78)

Then either set the default configuration to xbox or to h2v vanilla values, that including the crosshair offset.
Also add the missing upper FOV limit that was removed for some reason and split this in multiple commits. (one for removing the res fix leftovers, etc.)

Berthalamew:
Since you brought it up the crosshair offset shouldn't even be 0.138 in the first place. Vista by default uses 0.165, the same as original xbox.

Berthalamew:
Reverted back to 0.138 since even though the value is the same, the actual in-game offset is different. 
0.138 mimics the value on h2 xbox.

xlive/Blam/Engine/input/controllers.h
```diff
@@ -0,0 +1,11 @@
+#pragma once
+
+#define k_number_of_controllers 4
+
+enum e_controller_index : int
+{
+	controller_index_0 = 0,
```
nukeulater, line 7: I don't see the point of this enum

xlive/Blam/Engine/Networking/Session/NetworkSession.h
```diff
@@ -231,10 +233,10 @@ CHECK_STRUCT_SIZE(s_player_identifier, 8);
 
 struct s_session_interface_user
 {
-	char user_exists;
+	bool user_exists;
 	s_player_identifier network_user_identifier;
 	BYTE gap_B[3];
-	DWORD controller_index;
+	e_controller_index controller_index;
```
nukeulater, line 239: just do `int`

xlive/H2MOD/Modules/PlayerRepresentation/PlayerRepresentation.cpp
```diff
@@ -112,7 +113,12 @@ namespace PlayerRepresentation
 		else
 		{
 			if (H2Config_spooky_boy && !Memory::IsDedicatedServer())
-				*Memory::GetAddress<e_character_type*>(0x51A67C) = character_type_skeleton;
+			{
+				for (byte user_index = 0; user_index <= k_number_of_users; ++user_index)
```
nukeulater: out of bounds...

nukeulater: also doesn't follow the existing coding style

`for (int <name>; <name> < <condition>; <name>++)`

Berthalamew: Forgot to change the <= to a < when swapping from the controller index enum to user indexes in these loops. 
Fixed

PR #435, opened 2023-06-10
Fix crosshair scaling and update tag definitions

Berthalamew:
- cleanup player struct (needed for below)
- update s_new_hud_engine_globals

- remove unnecessary hook for turning off the hud
- added c_function_definition class
- update new_hud_definition tag definition
- update bitmap tag definition


xlive/H2MOD.cpp
```diff
@@ -58,6 +59,11 @@ std::unique_ptr<H2MOD> h2mod(std::make_unique<H2MOD>());
 bool H2XFirerateEnabled = false;
 bool xboxTickrateEnabled = false;
 
+bool get_xbox_tickrate_bool()
```
nukeulater: `xbox_tickrate_is_enabled()`

Berthalamew: changed

PR #445, opened 2023-08-26
add map_name commands

Berthalamew:
- added other commands to set local player counts, difficulty, game variant, and game mode

### DIFF
--- a/xlive/Blam/Engine/main/main_game.cpp
+++ b/xlive/Blam/Engine/main/main_game.cpp
@@ -36,9 +36,13 @@ void main_game_change(const s_game_options* options)
     return;
 }
 
-void game_set_difficulty(short difficulty)
+void main_game_launch_set_difficulty(short difficulty)
 {
-    if (difficulty < 4)
+    if (difficulty >= k_campaign_difficulty_levels_count || difficulty < 0)
+    {
+        LOG_ERROR_GAME("main_game_launch_set_difficulty: invalid difficulty {} (must be from 0-{})", difficulty, k_campaign_difficulty_levels_count - 1);
+    }
+    else
     {
         g_main_game_launch_options.difficulty = difficulty;
         g_main_game_launch_options.game_mode = _game_mode_campaign;
@@ -48,7 +52,11 @@ void game_set_difficulty(short difficulty)
 
 void main_game_launch_set_coop_player_count(int player_count)
 {
-    if (player_count <= k_number_of_users)
+    if (player_count > k_number_of_users || player_count < 1)
+    {
+        LOG_ERROR_GAME("main_game_launch_set_coop_player_count: invalid player count {} (must be from 1-{})", player_count, k_number_of_users);
+    }
+    else
     {
         g_main_game_launch_options.game_mode = _game_mode_campaign;
         g_main_game_launch_options.coop = player_count > 1;
@@ -59,7 +67,11 @@ void main_game_launch_set_coop_player_count(int player_count)
 
 void main_game_launch_set_multiplayer_splitscreen_count(int player_count)
 {
-    if (player_count <= k_number_of_users)
+    if (player_count > k_number_of_users || player_count < 1)
+    {
+        LOG_ERROR_GAME("main_game_launch_set_multiplayer_splitscreen_count: invalid player count {} (must be from 1-{})", player_count, k_number_of_users);
+    }
+    else
     {
         g_main_game_launch_options.game_mode = _game_mode_multiplayer;
         g_main_game_launch_user_count = player_count;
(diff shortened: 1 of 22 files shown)


xlive/Blam/Engine/main/main_game.cpp
```diff
@@ -0,0 +1,213 @@
...
+#include "Blam/Engine/cseries/cseries.h"
+#include "Blam/Engine/cseries/cseries_strings.h"
+#include "Blam/Engine/game/game.h"
+#include "Blam/Engine/Networking/Session/NetworkSession.h"
+#include "Blam/Engine/saved_games/game_variant.h"
+
+#include "Util/Hooks/Hook.h"
+
+s_game_options g_main_game_launch_options;
+DWORD g_main_game_launch_user_count;
```
nukeulater: Nit int

xlive/Blam/Engine/main/main_game.cpp
```diff
@@ -0,0 +1,213 @@
...
+    typedef void(__stdcall* main_game_change_t)(const s_game_options*);
+    auto p_main_game_change = Memory::GetAddress<main_game_change_t>(0x89BA, 0x1E4EC);
+    p_main_game_change(options);
+    return;
+}
+
+void main_game_launch_set_difficulty(short difficulty)
+{
+    if (!VALID_INDEX(difficulty, k_campaign_difficulty_levels_count))
+    {
```
nukeulater, line 49: Just return here if validation fails and remove the else statement. Also do it for the other similar cases

xlive/Blam/Engine/main/main_game.cpp
```diff
@@ -0,0 +1,213 @@
...
+    return;
+}
+
+void main_game_launch_setup_game_mode_details(void)
+{
+    switch (g_main_game_launch_options.game_mode)
+    {
+    case _game_mode_campaign:
+    {
+        if (g_main_game_launch_options.difficulty >= 0)
```
nukeulater: Use the range macros you just added.

xlive/H2MOD/GUI/imgui_integration/Console/CommandsUtil.h
```diff
@@ -193,25 +193,31 @@ class CircularStringBuffer
 		va_list valist;
 		va_start(valist, fmt);
 		int buffer_size_needed = _vsnprintf(NULL, 0, fmt, valist) + 1;
-		if (buffer_size_needed < m_line_buf_size)
+		if (buffer_size_needed != -1)
```
nukeulater: Will be -1 only when the result is -2. Practically never, because that function only returns - 1 if the buffer is insufficient.

xlive/H2MOD/GUI/imgui_integration/Console/CommandCollection.cpp
```diff
@@ -769,3 +776,56 @@ void CommandCollection::DeleteObject(datum objectDatumIdx)
 		Engine::Objects::object_destroy(objectDatumIdx);
 	}
 }
+
+int CommandCollection::map_name(const std::vector<std::string>& tokens, ConsoleCommandCtxData cbData)
+{
+	std::string name(tokens[1].begin(), tokens[1].end());
+	main_game_launch(name.c_str());
```
nukeulater, line 783: Why not just access the token directly, in the argument list, instead of creating another string?

nukeulater (changes requested):

xlive/H2MOD/GUI/imgui_integration/Console/CommandCollection.cpp
Berthalamew: Stack corruption happens when you try and do that on release builds

nukeulater: How could that cause stack corruption?

Berthalamew: No clue, something to do with cleanup after the map_name function is called from the console

xlive/H2MOD/GUI/imgui_integration/Console/CommandsUtil.h
Berthalamew: changed so it makes sure the value is positive, need this since we cast it to a size_t type afterwards and don't want it to cause undefined behaviour.

xlive/Blam/Engine/main/main_game.cpp
Berthalamew: changed

xlive/Blam/Engine/main/main_game.cpp
Berthalamew: the two globals are initialized now.

nukeulater: Meant you should make the variable int instead.

PR #451, opened 2023-09-24
add new crash report window and info dumps

Berthalamew:
- add utf8 datatype and unicode related functions
- add max_length member function
- add compress_file_to_zip function

~~Header bitmap for the window will need to be added in: "\\mods\\images\\halo2logo.bmp"~~
Header bitmap is now included in the dll as a resource

Image of the window: 
Capture


xlive/XLive/xbox/xbox.cpp
```diff
@@ -2,7 +2,7 @@
 
 #include "xbox.h"
 
-#include "resource.h"
+#include "resources\\resource.h"
```
nukeulater: single backslash here

xlive/resources/Resource.aps
nukeulater, line 1: don't commit resource cache files

PR #467, opened 2023-11-11
organize math definitions and functions

Berthalamew:
- moves them out of the folder they're currently in and into the proper modules they would've been in
- rename datatypes to their correct names
- remove operator overloads and use the correct math functions instead


xlive/Blam/Engine/interface/first_person_camera.cpp
```diff
@@ -60,15 +60,15 @@ float __cdecl player_control_get_field_of_view(int controller_index)
...
+			fov = DEGREES_TO_RADIANS(currentVariantSettings.forced_fov);
 		}
 		else if (player_control_fov_overridden)
 		{
 			fov = overridden_fov_radians;
 		}
 		else
 		{
-			fov = tags::get_tag_fast<s_unit_group_definition>(object_get_fast_unsafe(player_control_info->unit_datum_index)->tag_definition_index)->camera_field_of_view.as_rad();
+			fov = tags::get_tag_fast<s_unit_group_definition>(object_get_fast_unsafe(player_control_info->unit_datum_index)->tag_definition_index)->camera_field_of_view;
```
Kantanomo, line 71: missing conversion to radian.

xlive/Blam/Engine/math/math.h
```diff
@@ -0,0 +1,20 @@
...
+
+// Returns the minimum of the two values
+#define MIN(a, b) ((a) > (b) ? (b) : (a))
+
+// decimal part <0.5, floor
+// decimal part >0.5, ceil
+// TODO rework
+inline static auto blam_ticks_real_to_integer(float ticks_real) {
+	auto tick_val_integer = static_cast<int>(ticks_real + (ticks_real < 0.0f ? -0.5f : 0.5f));
+	return tick_val_integer;
```
Kantanomo, line 19: https://cplusplus.com/reference/cmath/round/
?

xlive/Blam/Engine/interface/first_person_camera.cpp
Berthalamew: The data is already in radians

xlive/Blam/Engine/math/math.h
Berthalamew: not sure if that's the same logic since if ticks_real is negative it subtracts -0.5f from itself then casts to an int

PR #469, opened 2023-11-18
add first person definitions + other misc additions

Berthalamew:
- update object flag names and add object_get_center_of_mass function
- update early mover struct
- add static flags class

### DIFF
--- a/xlive/Blam/Engine/cseries/cseries.h
+++ b/xlive/Blam/Engine/cseries/cseries.h
@@ -105,10 +105,10 @@ static_assert(sizeof(datum) == 4);
 #define SWAP_FLAG(flags, bit)			( (flags) ^=FLAG(bit) )
 #define FLAG_RANGE(first_bit, last_bit)	( (FLAG( (last_bit)+1 - (first_bit) )-1) << (first_bit) )
 
-#define BIT_VECTOR_SIZE_IN_LONGS(BIT_COUNT) (((BIT_COUNT) + (LONG_BITS - 1)) >> 5)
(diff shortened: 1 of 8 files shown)


xlive/Blam/Engine/cseries/cseries.h
```diff
@@ -100,6 +105,11 @@ static_assert(sizeof(datum) == 4);
 #define SWAP_FLAG(flags, bit)			( (flags) ^=FLAG(bit) )
 #define FLAG_RANGE(first_bit, last_bit)	( (FLAG( (last_bit)+1 - (first_bit) )-1) << (first_bit) )
 
+#define BIT_VECTOR_SIZE_IN_LONGS(BIT_COUNT) (((BIT_COUNT) + (LONG_BITS - 1)) >> 5)
+#define BIT_VECTOR_SIZE_IN_BYTES(BIT_COUNT) (4 * BIT_VECTOR_SIZE_IN_LONGS(BIT_COUNT))
+#define BIT_VECTOR_TEST_FLAG(BIT_VECTOR, BIT) (TEST_BIT(BIT_VECTOR[BIT >> 5], (BIT & (LONG_BITS - 1))))
```
nukeulater: `>> 5` is division by 32, `bit / 32` would be more clear, also I think this might not work for storage types smaller than 32 bits. [This code that's already implemented and available is better than this.](https://github.com/pnill/cartographer/blob/1f5cd0b3552ee73cf492bcb912e048f0cf349b57/xlive/Blam/Engine/memory/data.h#L6)

Berthalamew: These bitvector macros are only used for the c_static_flags_no_init class,
It's used for flags that need to be bigger than 32 bits

There's also c_flags that's a replacement for s_bitflags

PR #470, opened 2023-11-25
update tag definitions

Berthalamew:
- item, weapon, and render model updated
- update weapon_datum struct


xlive/Blam/Engine/items/weapon_definitions.cpp
```diff
@@ -0,0 +1,38 @@
+#include "stdafx.h"
+#include "weapon_definitions.h"
+
+weapon_first_person_interface_definition* first_person_interface_definition_get(_weapon_definition* definition, e_character_type character_type)
+{
+	weapon_first_person_interface_definition* result = NULL;
+	int32 index = 0;
+	switch (character_type)
```
Kantanomo: messed up indenting?

xlive/Blam/Engine/items/weapon_definitions.h
```diff
@@ -0,0 +1,519 @@
...
+    real32 runtime_ejection_port_recovery_rate;
+    real32 runtime_rate_of_fire_acceleration_rate;
+    real32 runtime_rate_of_fire_deceleration_rate;
+    real32 runtime_error_acceleration_rate;
+    real32 runtime_error_deceleration_rate;
+    tag_block<barrel_firing_effect> firing_effects; // firing effects determine what happens when this trigger is fired
+};
+TAG_BLOCK_SIZE_ASSERT(weapon_barrel_definition, 236);
+
+struct _weapon_definition
```
Kantanomo, line 398: maybe we should take some liberties in renaming things like this so it's easier to figure out what exactly we are looking at? idk just a thought I know we have _device_definition too now.

xlive/Blam/Engine/items/weapon_definitions.cpp
Berthalamew: fixed

xlive/Blam/Engine/items/weapon_definitions.h
Berthalamew: most of the definitions go by that naming scheme so there shouldn't be much confusion. See _item_definition

PR #474, opened 2023-12-10
add new unit definition

Berthalamew:


xlive/Blam/Engine/units/units.h
```diff
@@ -14,7 +20,7 @@ enum e_weapon_addition_method : int16
 	_weapon_addition_method_seven = 7,
 };
 
-enum e_unit_grenade_type : int8
+enum e_unit_grenade_type : int16
```
Kantanomo, line 23: could have uninteded consequences if it's used in other parts of the project, I.E add grenades etc

Berthalamew: It's only used for function parameters we've written ourselves. There shouldn't be any unintended consequences

PR #477, opened 2023-12-12
Enable function level linking

Berthalamew:


Berthalamew:
This removes around 2 thousand unreferenced functions

PR #549, opened 2024-02-04
add test DLL build text, including git commit hash and branch

nukeulater:
- easier to determine which DLL is the user using from videos

### DIFF
--- a/xlive/Blam/Engine/render/render_cartographer_ingame_ui.cpp
+++ b/xlive/Blam/Engine/render/render_cartographer_ingame_ui.cpp
@@ -3,6 +3,7 @@
 
 #include "Blam/Engine/game/game.h"
 #include "Blam/Engine/rasterizer/rasterizer_globals.h"
+#include "Blam/Engine/interface/hud.h"
 #include "Blam/Engine/shell/shell_windows.h"
 #include "Blam/Engine/text/draw_string.h"
 #include "Blam/Engine/text/font_cache.h"
@@ -12,6 +13,8 @@
 #include "H2MOD/Modules/Achievements/Achievements.h"
 #include "H2MOD/Modules/Shell/Config.h"
 
+#include "version_git.h"
+
 // define this to enable queueing a test message in render_cartographer_achievements
 // #define ACHIVEMENT_RENDER_DEBUG_ENABLED
 
@@ -19,6 +22,8 @@ const int32 k_status_text_font = 0;
 
 // defined in XLiveRendering.cpp
 extern char* buildText;
+
+bool cartographer_test_build_draw_text = true;
 
 void render_cartographer_status_text()
 {
@@ -41,9 +46,12 @@ void render_cartographer_status_text()
 	}
 
 	int32 line_height = get_text_size_from_font_cache(k_status_text_font);
-	if (game_is_main_menu || (!game_is_main_menu && paused_or_in_menus)) {
+	bool paused_or_in_main_menu = game_is_main_menu || paused_or_in_menus;
+	if (paused_or_in_main_menu) {
 		draw_string_reset();
 		draw_string_set_font_and_options(k_status_text_font, 0, 0, 0, &text_color_console, global_real_argb_black, false);
+
+		bounds.bottom = bounds.top + line_height;
 
 		if (H2Config_anti_cheat_enabled)
 		{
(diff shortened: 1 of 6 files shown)


punpckhdq:
something i wanted to do in the future was hook into the games build tracker system ("untracked version" etc) and populate it with git information. as well as remaking the overlay text certain halo 2 builds had

xlive/Blam/Engine/render/render_cartographer_ingame_ui.cpp
```diff
@@ -54,8 +62,47 @@ void render_cartographer_status_text()
...
 		bounds.top += line_height;
+		bounds.bottom = bounds.top + line_height;
 		draw_string_render(&bounds, master_state_string_buffer);
 	}
+
+#if defined(GEN_GIT_VER_VERSION_STRING)
+	if (cartographer_test_build_draw_text)
+	{
+		rasterizer_get_frame_bounds(&bounds);
+		int32 test_build_font = 0;
```
punpckhdq: i would use k_status_text_font here

xlive/Blam/Engine/render/render_cartographer_ingame_ui.cpp
```diff
@@ -54,8 +62,47 @@ void render_cartographer_status_text()
...
+
+		wchar_t result_text_buffer[1024];
+
+		utf8_string_to_wchar_string(GEN_GIT_VER_VERSION_STRING, build_git_version_string, NUMBEROF(build_git_version_string));
+		utf8_string_to_wchar_string(__DATE__, build_date, NUMBEROF(build_date));
+		utf8_string_to_wchar_string(__TIME__, build_time, NUMBEROF(build_time));
+		utf8_string_to_wchar_string(GET_GIT_VER_USERNAME, git_username, NUMBEROF(git_username));
+		utf8_string_to_wchar_string(GET_GIT_VER_BRANCH, git_branch_name, NUMBEROF(git_branch_name));
+
+		swprintf(result_text_buffer, NUMBEROF(result_text_buffer), L"%s %s", build_date, build_time);
```
punpckhdq: should this be one char buffer and one wchar buffer instead?

punpckhdq (commented):
since this is adding onto my code i'd thought i'd review it

punpckhdq:
one last comment: the draw_string functions have a cousin the actual ui code uses. it has proper scaling support iirc, maybe in the future that could be investigated

xlive/Blam/Engine/render/render_cartographer_ingame_ui.cpp
nukeulater: not sure I understand the question. If you're asking if it should be just 2 string buffers for each type of string encoding, it is possible, but easier to modify and play around with the code in this form.

xlive/Blam/Engine/render/render_cartographer_ingame_ui.cpp
nukeulater: ye

xlive/Blam/Engine/render/render_cartographer_ingame_ui.cpp
nukeulater: we need a vaargs variant of draw string tbh

punpckhdq: yeah i was confused about keeping all those ascii buffers. i thought about it again, shouldn't you be able to to use just one swprintf each instead of converting between ascii and wchar in your code at all? by using %S in your format string.

nukeulater: tried that and it didn't work for some reason, this is why I added those in the first place.

nukeulater: image
even MSVC complaining about it

punpckhdq: use a capital %S and not a %s in wide string formatting to signify an ascii string as a format parameter, https://learn.microsoft.com/en-us/cpp/c-runtime-library/format-specification-syntax-printf-and-wprintf-functions?view=msvc-170#type-field-characters

nukeulater: yeah that worked, thanks

PR #550, opened 2024-02-05
misc changes

Berthalamew:
- fix cloth attachment interpolation

### DIFF
--- a/xlive/Blam/Engine/objects/objects.cpp
+++ b/xlive/Blam/Engine/objects/objects.cpp
@@ -988,14 +988,17 @@ int16 __cdecl object_get_markers_by_string_id(datum object_index, string_id mark
 	return internal_object_get_markers_by_string_id(object_index, marker, marker_object, count, false);
 }
 
-// Replace calls to internal_object_get_markers_by_string_id
-void internal_object_get_markers_by_string_id_replace_calls(void)
+// Replace calls to functions getting the marker by string id
+void object_get_markers_by_string_id_replace_calls(void)
 {
 	// PatchCall(Memory::GetAddress(0x132792, 0x121662), internal_object_get_markers_by_string_id);	(Disabled as this is not interpolated)	TODO: Add parameter to internal_object_get_markers_by_string_id instead of not patchcalling
 	// PatchCall(Memory::GetAddress(0x1327B1, 0x121681), internal_object_get_markers_by_string_id);	(Disabled as this is not interpolated)	TODO: Add parameter to internal_object_get_markers_by_string_id instead of not patchcalling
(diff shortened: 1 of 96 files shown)


xlive/Blam/Engine/objects/widgets/widgets.cpp
```diff
@@ -1,7 +1,20 @@
...
 
+#include "main/interpolator.h"
+#include "objects/objects.h"
+
+
+
 void __cdecl widgets_new(datum object_index)
 {
 	INVOKE(0x14FFE2, 0x14CD7A, widgets_new, object_index);
+	object_initialize_for_interpolation(object_index);
```
nukeulater: why not just move this call to a later point in object_new? cause that's where widgets_new is getting called, and you're now initializing the interpolation data twice

Berthalamew: Forgot to remove this, the call is useless

xlive/Blam/Engine/objects/objects.cpp
```diff
@@ -988,14 +988,17 @@ int16 __cdecl object_get_markers_by_string_id(datum object_index, string_id mark
...
+void object_get_markers_by_string_id_replace_calls(void)
 {
 	// PatchCall(Memory::GetAddress(0x132792, 0x121662), internal_object_get_markers_by_string_id);	(Disabled as this is not interpolated)	TODO: Add parameter to internal_object_get_markers_by_string_id instead of not patchcalling
 	// PatchCall(Memory::GetAddress(0x1327B1, 0x121681), internal_object_get_markers_by_string_id);	(Disabled as this is not interpolated)	TODO: Add parameter to internal_object_get_markers_by_string_id instead of not patchcalling
-	PatchCall(Memory::GetAddress(0x134C26, 0x123AF6), internal_object_get_markers_by_string_id);
 	// PatchCall(Memory::GetAddress(0x13823D, 0x12710D), internal_object_get_markers_by_string_id);	(Disabled as this is not interpolated)	TODO: Add parameter to internal_object_get_markers_by_string_id instead of not patchcalling
 	// PatchCall(Memory::GetAddress(0x138257, 0x1214CB), internal_object_get_markers_by_string_id);	(Disabled as this is not interpolated)	TODO: Add parameter to internal_object_get_markers_by_string_id instead of not patchcalling
+	PatchCall(Memory::GetAddress(0x134C26, 0x123AF6), internal_object_get_markers_by_string_id);	// object_animation_callback
```
nukeulater: this line hooks something audio related rather than visual, probably not even called outside game tick context

Berthalamew: It's used for an animation callback function, from what I can tell it has to do with animation events and it creates a sound at that position

PR #622, opened 2024-04-03
main game time updates

nukeulater:
- and some other minor changes

### DIFF
--- a/xlive/Blam/Engine/effects/particle_system.h
+++ b/xlive/Blam/Engine/effects/particle_system.h
@@ -38,7 +38,7 @@ CHECK_STRUCT_SIZE(s_particle_system_update, 0xC);
 class c_particle_system
 {
 public:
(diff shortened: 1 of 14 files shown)


xlive/Blam/Engine/effects/particle_system.h
```diff
@@ -38,7 +38,7 @@ CHECK_STRUCT_SIZE(s_particle_system_update, 0xC);
 class c_particle_system
 {
 public:
-	int32 datum_salt;
+	uint16 datum_salt;
```
Berthalamew, line 41: Should we define padding below this so that it's explicit?

Berthalamew (approved):

xlive/Blam/Engine/effects/particle_system.h
nukeulater: not necessarily

PR #624, opened 2024-04-07
Allow proper multi-controller support in H2V

Himanshu-01:
# Includes
* Adds `input_globals` and `input_abstraction_globals`
* Fixes some of the hired-gun patches that tied all inputs to `input_globals.gamepad_state[0]`
* User interfaces like pregame lobby can now detect multiple controller inputs successfully

# Todo List
* Completely breaks Imgui AdvancedSetting ControllerInput.cpp hacky patches that will crash you anytime
* `input_abstraction_update()` needs to be redone to fix all other issues
* Certain UIs like main_menu dont like multiple local users

### DIFF
--- a/xlive/Blam/Engine/game/player_vibration.cpp
+++ b/xlive/Blam/Engine/game/player_vibration.cpp
@@ -12,9 +12,9 @@
 #include "H2MOD/Tags/TagInterface.h"
 
 s_vibration_globals* vibration_globals_get(void);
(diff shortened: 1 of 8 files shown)


Halo2Guy:
Lookin Good Gentlemen. 

Keep up the hard work everyone 🏆😎🏆 

#Halo2ProjectCartographerBest

PR #626, opened 2024-04-15
Cleanup and warning fixes

Berthalamew:
- update c_flags to use c_flags_no_init as a base class
- add asserts
- fix warnings/errors on non-msvc compilers
- add msvc macro

### DIFF
--- a/xlive/Blam/Engine/cseries/cseries.h
+++ b/xlive/Blam/Engine/cseries/cseries.h
@@ -53,7 +53,7 @@ static_assert(sizeof(datum) == 4);
 #define LONG_BITS SIZEOF_BITS(int32)
 
 #if(!__INTEL_LLVM_COMPILER)
-#define MSVC
+#define IS_MSVC_COMPILER
 #endif
 
 // Invokes a function
--- a/xlive/Blam/Engine/cseries/cseries.h
+++ b/xlive/Blam/Engine/cseries/cseries.h
@@ -52,6 +52,10 @@ static_assert(sizeof(datum) == 4);
 #define SHORT_BITS SIZEOF_BITS(int16)
 #define LONG_BITS SIZEOF_BITS(int32)
 
+#if(!__INTEL_LLVM_COMPILER)
+#define MSVC
+#endif
+
 // Invokes a function
 // ADDR_CLIENT: file offset in halo2.exe
 // ADDR_SERVER: file offset in h2server.exe
--- a/xlive/Blam/Engine/cseries/cseries_strings.cpp
+++ b/xlive/Blam/Engine/cseries/cseries_strings.cpp
@@ -1,22 +1,42 @@
 #include "stdafx.h"
 #include "cseries_strings.h"
 
-size_t csstrnlen(const char* string, size_t size)
+size_t csstrnlen(const char* s, size_t size)
 {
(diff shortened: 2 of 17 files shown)


xlive/Blam/Engine/main/main_game.cpp
```diff
@@ -51,7 +51,11 @@ void main_game_launch_set_coop_player_count(int32 player_count)
 {
     if (!IN_RANGE_INCLUSIVE(player_count, 1, k_number_of_users))
     {
+#ifdef MSVC
         error(2, __FUNCTION__": invalid player count %d (must be from 1-%d)", player_count, k_number_of_users);
+#else
+        error(2, "%s: invalid player count %d (must be from 1-%d)", __FUNCTION__, player_count, k_number_of_users);
```
nukeulater, line 54: I don't see why this wouldn't work using MSVC

xlive/Blam/Engine/memory/static_arrays.h
```diff
@@ -41,55 +46,80 @@ class c_flags
...
+	c_flags_no_init<t_type, t_storage_type, k_count> operator&(c_flags_no_init<t_type, t_storage_type, k_count>& value) const
 	{
-		return ~m_storage;
+		c_flags_no_init<t_type, t_storage_type, k_count> result;
+		result.set_unsafe(m_storage & value.m_storage);
+		return result;
 	}
 
-	bool operator==(const c_flags<t_type, t_storage_type, k_count>& value)
+	void operator|=(c_flags_no_init<t_type, t_storage_type, k_count>& value)
```
nukeulater, line 85: this should also return

xlive/Blam/Engine/cseries/cseries_strings.cpp
```diff
@@ -1,22 +1,42 @@
...
+int32 cvsnzprintf(char* buffer, size_t size, char const* format, ...)
+{
+	ASSERT(buffer);
+	ASSERT(format);
+	ASSERT(size > 0);
+
+	va_list va_args;
+	va_start(va_args, format);
+	int32 result = vsnprintf(buffer, size, format, va_args);
+	buffer[size - 1] = 0;
```
nukeulater: nitpicking: `= '\0'`

xlive/Blam/Engine/cseries/cseries.h
```diff
@@ -52,6 +52,10 @@ static_assert(sizeof(datum) == 4);
 #define SHORT_BITS SIZEOF_BITS(int16)
 #define LONG_BITS SIZEOF_BITS(int32)
 
+#if(!__INTEL_LLVM_COMPILER)
+#define MSVC
```
nukeulater: this name specifier is quite generic, might be conflicting with existing directives

xlive/Blam/Engine/memory/static_arrays.h
nukeulater: actually nvm, perhaps this operator shouldn't be allowed to be chained in the same expression.

xlive/Blam/Engine/main/main_game.cpp
Berthalamew: Original string appends the function name to the string while this isn't possible on non-msvc compilers.

PR #640, opened 2024-05-16
Fixing a Bertha

Kantanomo:
undoing using a define out of place that worked coincidentally.


twist84:
It may be better to have a username length/size constant

Berthalamew:
use the XUSER_NAME_SIZE macro

PR #660, opened 2024-07-15
fix network adapter crash + other misc changes

Berthalamew:


Berthalamew:
Confirmation of testing on dedi after changes:
image

PR #664, opened 2024-07-28
Fix assert in matrix3x3_rotation_to_quaternion

Berthalamew:


Berthalamew:
The offending assert:
image

PR #677, opened 2024-08-12
add host control over the player's team using session membership, network session updates for eventual refactor

nukeulater:


xlive/Blam/Engine/networking/NetworkMessageTypeCollection.cpp
```diff
@@ -283,20 +283,29 @@ void __stdcall handle_channel_message_hook(void* thisx, int network_channel_inde
...
 			&& peer_network_channel->get_network_address(&addr))
 		{
-			auto peer_index = NetworkSession::GetPeerIndexFromNetworkAddress(&addr);
+			int32 peer_index = NetworkSession::GetPeerIndexFromNetworkAddress(&addr);
 			EventHandler::NetworkPlayerEventExecute(EventExecutionType::execute_after, peer_index, EventHandler::NetworkPlayerEventType::add);
 			NetworkMessage::SendAntiCheat(peer_index);
 		}
+		break;
+	case k_network_message_type_collection_count:
+	default:
```
Berthalamew, line 300: remove the count case and add the following to the default case:
DISPLAY_ASSERT("unreachable")

Berthalamew (changes requested):

xlive/Blam/Engine/networking/NetworkMessageTypeCollection.cpp
nukeulater: no, because it shouldn't throw an error. Going to remove the unnecessary cases

Berthalamew: ah my bad, realized it's a hook and not the actual switch case.

PR #681, opened 2024-08-20
add assertions and error logging for object functions

Berthalamew:
- remove unused HS functions


xlive/Blam/Engine/tag_files/tag_files.cpp
```diff
@@ -4,4 +4,12 @@
...
 	return INVOKE(0x30D58, 0x24C08, tag_files_open);
-}
\ No newline at end of file
+}
+
+const char* tag_name_strip_path(const char* name)
+{
+    ASSERT(name);
+
+    const char* result = strrchr(name, 92);
```
nukeulater, line 13: ascii code instead of the actual character?

PR #685, opened 2024-08-27
fix for rewritten tag_get_unsafe and injected tags

Kantanomo:
the rewritten tag_get_unsafe was rewritten to use a different pointer to the tag_header->tag_instances. The pointer it originally used was a on-compilation pointer added by the inlining of tag getting operations.

Updated the code of the injector to also point the g_cache_file_memory_globals->tag_header.tag_instances to point to the redirected tag_instance table.


nukeulater:
that fixed it

PR #695, opened 2024-09-06
adjust header includes

Berthalamew:
- improve build time by removing header dependencies and lowering amount of times certain headers are parsed count when compiling
- remove static string usage from some structures
- move k_number_of_users to players
- remove VariantMPGameEngine from build (never been used but we'll keep it for now)
- move e_game_team to game_allegiance


Berthalamew:
Trace before changes:
trace_before

Trace after changes:
trace_after

PR #700, opened 2024-09-07
infection team change switch for split screen fix

Kantanomo:


xlive/H2MOD/Variants/Infection/Infection.cpp
```diff
@@ -95,14 +95,18 @@ void Infection::sendTeamChange()
 					{
 						if (!Memory::IsDedicatedServer())
 						{
-							user_interface_controller_set_desired_team_index(_controller_index_0, team);
-							user_interface_controller_update_network_properties(_controller_index_0);
+							player_indexes[player_array_index] = i;
```
nukeulater: this is not really needed for local players, just the interface code

nukeulater: also https://github.com/pnill/cartographer/blob/8e566ef2f8bded4e323536fcb00b69174f1a9402/xlive/H2MOD/Variants/Infection/Infection.cpp#L467 needs to be taken care of

Kantanomo: it is needed or the session doesn't update properly, the local profile will change teams but any other players will still see them on the red team

nukeulater: that shouldn't be an issue, there must be something else wrong

Kantanomo: from all my testing it wont work without it.

Kantanomo: removed

PR #704, opened 2024-09-14
map manager updates and cleanup

nukeulater:
updated some other components used by the map manager

### DIFF
--- a/xlive/H2MOD/Modules/MapManager/MapManager.cpp
+++ b/xlive/H2MOD/Modules/MapManager/MapManager.cpp
@@ -9,64 +9,42 @@
 #include "networking/NetworkMessageTypeCollection.h"
 #include "networking/Session/NetworkSession.h"
 
+#include "networking/logic/life_cycle_manager.h"
+#include "main/game_preferences.h"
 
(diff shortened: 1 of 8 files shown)


xlive/H2MOD/Modules/MapManager/MapManager.cpp
```diff
@@ -255,55 +233,49 @@ char __cdecl handle_map_download_callback() {
...
 const wchar_t* get_receiving_map_string()
 { 
-	int(__cdecl* get_default_game_language)() = (int(__cdecl*)())((char*)Memory::GetAddress(0x381fd));
-	wchar_t** receiving_map_message = Memory::GetAddress<wchar_t**>(0x46575C);
+	e_language current_language = get_current_language();
+	wchar_t** map_receiving_default_text = Memory::GetAddress<wchar_t**>(0x46575C);
 
-	if (get_default_game_language() == 0) // check if english
-		return receiving_map_string[0];
+	if (current_language == _language_english) // check if english
```
Berthalamew: This doesn't need to be here since we're getting the array from ingame regardless

Berthalamew (changes requested):

PR #708, opened 2024-09-29
add shader model 3 support and rewrite dx9 initialization functions to add d3d9on12 support

Berthalamew:
- update atmospheric fog apply shader to be partial precision
- reorder render calls in render_scene to draw sky later on (improving performance)
- fix condition in simulation_queue_entity_decode_header


xlive/Blam/Engine/render/render.cpp
```diff
@@ -304,6 +348,428 @@ void __cdecl render_window(window_bound* window, bool is_texture_camera)
...
+                *global_rasterizer_pixel_shader_index_get() = 2;
+                g_dx9_dont_draw_to_depth_target_if_mrt_is_used = false;
+            }
+
+            rasterizer_dx9_perf_event_begin("texaccum", NULL);
+            if (prepare_render_layer(_render_layer_texture_accumulate))
+            {
+                draw_render_layer();
+                reset_after_render_layer_draw();
+            }
```
nukeulater: function instead? this sequence is quite repetitive

Berthalamew: added the DRAW_RENDER_LAYER macro

PR #714, opened 2024-10-20
Rewrite screenshot_render + add screenshot_cubemap command + misc changes

Berthalamew:
- screenshot render function rewritten to make sure the screenshot render target is only created if we're taking a screenshot, rather than every frame
- screenshots are currently broken in split screen (legacy bug), we can fix it at some point in the future?
- screenshot_cubemap takes in the name of the tif file as a parameter which will be saved in the main game directory


xlive/Blam/Engine/cache/cache_files.cpp
```diff
@@ -217,7 +217,10 @@ bool __cdecl scenario_tags_load(const char* scenario_path)
 
 	const uint32 aligned_tag_size_read = cache_header->tag_size + cache_header->tag_offset_mask;
 
-	if(!cache_header_verify(cache_header) || strlen(cache_header->version_string) > 32)
+	if(!cache_header_verify(cache_header)
+		// This condition will never be true, included by bungie for some reason
```
Kantanomo: This is essentially a sanity check if, for example if a cache header seems correct but there is data that exists where a null terminator should be the strlen will be invalid.. also could maybe just be a part of cache_header_verify.

Berthalamew: readded

PR #717, opened 2024-11-01
add debug_render_horizontal_splitscreen command

Berthalamew:
- debug_render_horizontal_splitscreen forces the display split type to horizontal
- add intel sky hack to disable skybox rendering (newer intel cards crash when rendering the sky)
- rename s_camera to render_camera

### DIFF
--- a/xlive/Blam/Engine/render/render.h
+++ b/xlive/Blam/Engine/render/render.h
@@ -10,7 +10,7 @@
 
 /* enums */
 
-enum e_screen_split_type : int8
+enum e_screen_split_type : uint32
 {
 	_screen_split_type_full = 0,
 	_screen_split_type_half = 1,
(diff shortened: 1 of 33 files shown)


xlive/Blam/Engine/render/render.h
```diff
@@ -65,8 +71,8 @@ struct window_bound
 	int32 single_view;
 	int32 window_bound_index;
 	int32 user_index;
-	s_camera render_camera;
-	s_camera rasterizer_camera;
+	struct render_camera render_camera;
+	struct render_camera rasterizer_camera;
```
Kantanomo, line 75: ? Struct struct?

Berthalamew: Intended, does not compile unless you use struct in front of it.
These are the original member and struct names.

PR #733, opened 2024-11-08
use plugins folder instead of definitions + misc changes

Berthalamew:
- fix conflicting parameter for c_animation_manager::get_node_orientations
- add mask entries to e_object_type and replace usage in code
- change e_object_type to the correct size
- adjust names in object_datum struct
- rename s_object_header to object_header_datum and adjust names
- make get_object_table a private function
- update names in e_object_data_flags
- adjust names for e_unit_grenade_type
- adjust names in scenario_object_name
- move node_render_leaves to structures.h

### DIFF
--- a/xlive/Blam/Engine/objects/object_type_list.h
+++ b/xlive/Blam/Engine/objects/object_type_list.h
@@ -3,7 +3,7 @@
 // TODO move this into object_types.h but currently getting issues
 // Has to do with header recursion and errors that pop up with tag definitions?
 
-enum e_object_type : int8
+enum e_object_type : int32
 {
 	_object_type_biped = 0,
 	_object_type_vehicle = 1,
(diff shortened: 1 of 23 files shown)


twist84:
*finally*

xlive/Blam/Engine/objects/object_type_list.h
```diff
@@ -20,6 +20,31 @@ enum e_object_type : int8
 	_object_type_creature = 12,
 	k_object_types_count,
 
-	_object_type_flag_unit = FLAG(_object_type_biped) | FLAG(_object_type_vehicle)
+	_object_mask_all = NONE,
```
nukeulater, line 23: why not use another enum for these?

nukeulater: and keep the underlying type the same (int8)?

PR #739, opened 2024-11-09
fix crosshair scaling

Kantanomo:


xlive/H2MOD.cpp
```diff
@@ -390,6 +390,8 @@ bool __cdecl OnMapLoad(s_game_options* options)
 
 	MetaExtender::free_tag_blocks();
 	TagFixes::OnMapLoad();
+
+	hud_draw_on_map_load();
```
Berthalamew: We don't need to do this on dedis

Berthalamew (changes requested):

xlive/H2MOD.cpp
Kantanomo: fixed

xlive/Blam/Engine/interface/new_hud_draw.cpp
```diff
@@ -747,6 +758,32 @@ void __cdecl draw_hud_player_indicators(uint32 local_render_user_index)
 }
 /* public code */
 
+void hud_draw_on_map_load()
+{
+	if (g_draw_hud_crosshair_bitmap_cache)
+		free(g_draw_hud_crosshair_bitmap_cache);
```
Berthalamew, line 764: g_draw_hud_crosshair_bitmap_cache should also be freed when we close the game, not just here

Berthalamew (changes requested):

PR #750, opened 2024-11-20
color code updates

nukeulater:


xlive/Blam/Engine/math/color_math.cpp
```diff
@@ -1,25 +1,48 @@
 #include "stdafx.h"
 #include "color_math.h"
 
+const real_rgb_color global_real_rgb_color_table[] =
```
Berthalamew: This just duplicates the rgb data again.
The purpose of the color table was so they didn't have to duplicate data between the argb and rgb versions of the color.
The table should be left as-is.

Berthalamew (changes requested):

xlive/Blam/Engine/math/color_math.cpp
nukeulater: those casts in the header have to go though

PR #751, opened 2024-11-26
Misc changes

Berthalamew:
- add new tag definitions (sound class, item, user interface globals, crate, device control, device light fixture, device machine, equipment, and garbage)
- adjust grenade enums
- add memory/static_arrays.h to precompiled header
- cleanup map slots code
- refactor weapon levels to use an array instead of a map
- use sizeof in tag loader code
- use function pointers instead of std::function in imgui handler code
- remove prototypes for deleted functions in new_hud.cpp
- replace std::fstream usage with fprintf in Config.cpp and Accounts.cpp
- remove pragma pack and pop from input abstraction (not needed)
- adjust loops for crosshair resizing code
- adjust types for hud draw code
- move rasterizer_globals_get_data to game_globals
- add logging condition when scenario is unsupported (always printed it was unsupported before, regardless of version)


xlive/Blam/Engine/sound/sound_definitions.h
```diff
@@ -84,7 +42,7 @@ struct sound_definition_v1
 {
 	c_flags_no_init<e_sound_definition_flags, uint16, k_sound_definition_flag_count> flags;
 
-	e_sound_definition_class sound_class;
+	int8/*e_sound_class_type*/ sound_class;
```
Kantanomo, line 45: ?

Berthalamew: e_sound_class_type is the enum and it's an int32, they use int8 in the definition for padding reasons

twist84: You guys should add `c_enum` for this very reason

PR #753, opened 2024-11-30
add global real math constants + misc changes  

Berthalamew:
- simplify imgui string table logic
- fix motd downloading when using the portable setting (would previously download in localappdata)
- adjust logic for strings in add_button_key_split_input
- remove unneeded includes from screen_cartographer_errors.h and screen_cartographer_account_manager.h
- adjust define names in user_interface_text.h


Kantanomo:
👍

PR #760, opened 2024-12-07
add password confirmation to account creation dialog, use the xbox task dialog during account creation request

nukeulater:
- switched error menu type to error message with continue button, fixed the button handler when widgets are using the error dialog UI channel


nukeulater:
@Himanshu-01

PR #764, opened 2024-12-16
xliveless code changes

Berthalamew:
- trim xliveless code to remove logging for unused ordinals and remove code not relevant to h2v
- remove begin and end member functions from tag_block
- update tagblock [] overload to use tag_block_get_element_with_size
- disable C++ stack unwinding generation for every function
- make h2mod static and remove unused global
- remove unused global in new_hud
- update modern aiming notes in advanced settings imgui menu


nukeulater:
I don't see the point in removing the xlive code

Berthalamew:
We don't really need to keep the code around that isn't useful for h2.
ie. GTAIV checks and marketplace code + globals, etc...

PR #774, opened 2025-01-09
preliminary work for projection changes

Berthalamew:
changes and additions to prepare for projection work

### DIFF
--- a/xlive/Blam/Engine/math/matrix_math.cpp
+++ b/xlive/Blam/Engine/math/matrix_math.cpp
@@ -130,48 +130,45 @@ void matrix4x3_from_point_and_vectors(real_matrix4x3* matrix, const real_point3d
 	return;
 }
 
-void matrix4x3_inverse(const real_matrix4x3* input, real_matrix4x3* output)
+real_matrix4x3* matrix4x3_inverse(const real_matrix4x3* input, real_matrix4x3* output)
 {
-	if (input->scale == 0.0f)
-	{
-		csmemset(output, 0, sizeof(real_matrix4x3));
-	}
-	else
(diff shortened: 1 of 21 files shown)


xlive/Blam/Engine/math/matrix_math.cpp
```diff
@@ -130,48 +130,45 @@ void matrix4x3_from_point_and_vectors(real_matrix4x3* matrix, const real_point3d
 	return;
 }
 
-void matrix4x3_inverse(const real_matrix4x3* input, real_matrix4x3* output)
+real_matrix4x3* matrix4x3_inverse(const real_matrix4x3* input, real_matrix4x3* output)
```
nukeulater: checked this function, in this instance the out vector pointer is not actually returned

xlive/Blam/Engine/rasterizer/dx9/rasterizer_dx9.h
```diff
@@ -50,10 +50,9 @@ class c_rasterizer_constant_4f_cache
             {
                 if (count != i)
                 {
-                    for (int32 j = i + index; j < size; j++)
+                    for (i; i < count; ++i)
```
nukeulater: you can remove the first expression in this
i.e `for (; i < count; ++i)`

and `if (count != i)` could also be removed

Berthalamew: Changed

xlive/Blam/Engine/math/matrix_math.cpp
Berthalamew: removed the return value

PR #780, opened 2025-01-30
XNet, join code refactoring, scrapped the idea of single socket connections (for now)

nukeulater:
- ~~improved network port map handling, **now connection negotiation requires just a single port to be forwarded**~~
- ~~**moved the QoS port to base port as well**~~
- **scrapped the idea of single socket connections, this would require a lot more work than initially planned, but keep the code refactor and improvements in place**
- moved debug configuration to another config key. This will have the effect of disabling debug logs for people that may have forgotten to disable them after debugging, which will result in overall performance improvements

### DIFF
--- a/xlive/XLive/xnet/upnp.cpp
+++ b/xlive/XLive/xnet/upnp.cpp
@@ -50,13 +50,10 @@ void ForwardPorts()
 	ModuleUPnP upnp;
 	Utils::UPnPResult upnpResult(Utils::UPnPErrorType::None, 0);
 
-	upnpResult = upnp.UPnPForwardPort(false, H2Config_base_port, H2Config_base_port, "Halo2");
+	upnpResult = upnp.UPnPForwardPort(false, H2Config_base_port, H2Config_base_port, "Halo2_port");
 	LOG_INFO_NETWORK("ForwardPorts() - Halo2 port forwarding result: {}", upnpResult.ErrorCode);
 
-	upnpResult = upnp.UPnPForwardPort(false, (H2Config_base_port + 1), (H2Config_base_port + 1), "Halo2_1");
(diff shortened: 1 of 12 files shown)


xlive/XLive/xnet/upnp.cpp
```diff
@@ -50,14 +50,14 @@ void ForwardPorts()
...
 	LOG_INFO_NETWORK("ForwardPorts() - Halo2 port forwarding result: {}", upnpResult.ErrorCode);
 
-	upnpResult = upnp.UPnPForwardPort(false, (H2Config_base_port + 1), (H2Config_base_port + 1), "Halo2_1");
-	LOG_INFO_NETWORK("ForwardPorts() - Halo2_1 port forwarding result: {}", upnpResult.ErrorCode);
+	upnpResult = upnp.UPnPForwardPort(false, H2Config_base_port + 1, H2Config_base_port + 1, "Halo2_port1");
+	LOG_INFO_NETWORK("ForwardPorts() - Halo2 port forwarding result: {}", upnpResult.ErrorCode);
 
-	upnpResult = upnp.UPnPForwardPort(true, (H2Config_base_port + 10), (H2Config_base_port + 10), "Halo2_QoS");
-	LOG_INFO_NETWORK("ForwardPorts() - Halo2_QoSport forwarding result: {}", upnpResult.ErrorCode);
+	upnpResult = upnp.UPnPForwardPort(true, H2Config_base_port + 10, H2Config_base_port + 10, "Halo2_port_QoS");
```
Berthalamew, line 59: Should probably make the values of 1 and 10 here constants across the project. ie `H2Config_base_port + k_QoS_port_offset` or something similar

PR #796, opened 2025-02-13
fix language setting

Berthalamew:
- main_time re-org


nukeulater:
have other people test before merging, because I don't see any "slowdown", or at least anything noticeable or worse.

PR #816, opened 2025-03-10
string changes

Berthalamew:
- remove cseries_strings file
- misc type changes


xlive/Util/Hooks/Hook.cpp
```diff
@@ -84,46 +84,46 @@ void RetourClassFunc(BYTE *src, BYTE *restore, const unsigned int len)
...
 {
 	VirtualProtectAndExecutePatch((LPVOID)destAddress, numBytes, PAGE_EXECUTE_READWRITE,
 
 	memcpy((LPVOID)destAddress, bytesToWrite, numBytes);
 	);
 }
 
-void PatchCall(DWORD call_addr, DWORD new_function_ptr) {
-	DWORD callRelative = new_function_ptr - (call_addr + 5);
+void PatchCall(size_t call_addr, size_t new_function_ptr) {
```
nukeulater: ?

Berthalamew: using uintptr_t instead as we discussed before

PR #823, opened 2025-03-21
fix server instance numbers for config

Berthalamew:
- dedis will now use local configs by default
- dedis that have an instance name supplied will use h2serverconfig appended by the instance parameter supplied
- dedis will now use the instance name appended to the log instance folder
- add unicode helper functions
- replace Memory::IsDedicatedServer with shell_is_dedicated_server
- make server use our rewritten shell_initialize function
- remove RunLoop
- move crc table in Util header to cpp
- add custom setting struct in playlist loader
- remove experimental rendering enum
- remove filesys
- fix shadow and water graphics options defaulting to low
- don't initialize console commands on dedis


xlive/Blam/Engine/shell/shell.cpp
```diff
@@ -80,107 +96,79 @@ const char* shell_get_version(void)
...
 	cache_file_verification_initialize();
+	cseries_initialize();
 	runtime_state_initialize();
 
-	int arg_count;
-	wchar_t** cmd_line_args = LOG_CHECK(CommandLineToArgvW(GetCommandLineW(), &arg_count));
-	if (cmd_line_args && arg_count > 1) {
-		for (int i = 1; i < arg_count; i++) {
-			wchar_t* cmd_line_arg = cmd_line_args[i];
+	SYSTEM_DEBUG_MEMORY(cseries_initialize());
```
nukeulater, line 111: speaking of `SYSTEM_DEBUG_MEMORY`, the macro should also have a "before" initialization message, and all of these initialization calls should be 1 liners:

```
SYSTEM_DEBUG_MEMORY(real_math_initialize());
```

not:

```
real_math_initialize()
SYSTEM_DEBUG_MEMORY(real_math_initialize());
```

where SYSTEM_DEBUG_MEMORY is different depending on the compile configuration used

Berthalamew: I think this is actually how the macro was originally, since `runtime_state_initialize` is called after `cseries_initialize` but the debug memory internal call is done after `runtime_state_initialize`

Berthalamew: Same thing with `animation_cache_new`. There's no animation cache on halo 2 vista but a memory internal call is done after it would of been there like on xbox.

nukeulater: nvm, there's no "before" message, I miscalled that

Berthalamew: There does seem to be a before macro as well (i.e. before XNetStartup), they're just not used in that shell code.

PR #832, opened 2025-04-16
bug fixes

Berthalamew:
- fix drop logic
- fix bug where the console wouldn't display any output
- fix title on multiple game instances


xlive/H2MOD/GUI/imgui_integration/Console/CommandCollection.cpp
```diff
@@ -225,15 +225,15 @@ static int CommandCollection::BoolVarHandlerCmd(const std::vector<std::string>&
 	if (tokens.size() - 1 < 1)
 	{
 		booleanCmdVar->SetVal(!booleanCmdVar->GetVal());
-		outputCb(StringFlag_None, "%s", booleanCmdVar->AsString().c_str());
+		outputCb(StringFlag_None, booleanCmdVar->AsString().c_str());
```
nukeulater, line 228: why wouldn't this work anymore??

Berthalamew: It works fine either way.
Wouldn't we want to pass the format as the string itself?
Rather than passing a specifier for just a string and then the string?

nukeulater: below there are instances where the text has some spacing, so not sure why you would change this.

Berthalamew: Thought it was a mistake as it was using tab within the source code vs the tab character.
I'll change it to use the tab character.

nukeulater:
this text is not getting logged anymore:  [Reading H2Configuration file...](https://github.com/pnill/cartographer/pull/832/files#diff-76733ff73f98ccaf2c189ba693986c55a6509c1613f62d1020902ecd182056d0R569)

Berthalamew:
fixed

PR #853, opened 2025-05-15
fix bug where server_playlist config option wouldn't work

Berthalamew:
- implement shell argument commands in kablam
- remove instance name globals and call kablam functions instead
- create kablam_config for some config related code

### DIFF
--- a/xlive/Blam/Engine/kablam/kablam_log.cpp
+++ b/xlive/Blam/Engine/kablam/kablam_log.cpp
@@ -26,24 +26,28 @@ void kablam_log_apply_patches(void)
 
 void kablam_log_initialize(void)
 {
-	wchar_t log_name[MAX_PATH] = L"reports\\";
-
-	// Make sure reports folder exists
-	CreateDirectoryW(log_name, NULL);
-
-	const wchar_t* instance_name = kablam_shell_argument_get(L"-instance:");
-	if (instance_name)
+	// Only log if -log parameter is present
+	if (kablam_shell_argument_index_get(L"-log") != NONE)
 	{
-		// Make sure instance folder exists
-		ustrncat(log_name, instance_name, NUMBEROF(log_name));
-		ustrncat(log_name, L"\\", NUMBEROF(log_name));
+		wchar_t log_name[MAX_PATH] = L"reports\\";
+
+		// Make sure reports folder exists
 		CreateDirectoryW(log_name, NULL);
+
+		const wchar_t* instance_name = kablam_shell_argument_get(L"-instance:");
+		if (instance_name)
+		{
+			// Make sure instance folder exists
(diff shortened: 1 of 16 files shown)


xlive/Blam/Engine/networking/Session/network_session.h
```diff
@@ -61,6 +61,14 @@ enum e_network_session_mode
 	k_network_session_mode_count = 10
 };
 
+enum e_network_game_privacy : int32
+{
+	_network_game_privacy_open,
+	_network_game_privacy_invite,
+	_network_game_privacy_closed,
+	k_network_game_privacy_count
```
Kantanomo, line 70: gold only?

xlive/H2MOD/Modules/Shell/Startup/Startup.cpp
```diff
@@ -102,14 +103,16 @@ void InitLocalAppData()
...
 	// We set the instance string based on whether or not 
-	const wchar_t* instance_string;
+	const wchar_t* instance_string = NULL;
 	wchar_t instance_number_string[4];
-	if (config_use_instance_name())
+
+	if (config_use_instance_name(&instance_string))
 	{
-		instance_string =  g_shell_windows_instance_name;
+		// Do nothing, instance_string is set by the above function
```
Kantanomo: invert the if-else if no code is needed for if the condition is true

xlive/Blam/Engine/networking/Session/network_session.h
Berthalamew: There's no privacy setting for gold

Berthalamew: Nevermind there is, added

xlive/H2MOD/Modules/Shell/Startup/Startup.cpp
Berthalamew: changed

xlive/Blam/Engine/kablam/kablam_log.cpp
```diff
@@ -0,0 +1,68 @@
...
+	{
+		// Make sure instance folder exists
+		ustrncat(log_name, instance_name, NUMBEROF(log_name));
+		ustrncat(log_name, L"\\", NUMBEROF(log_name));
+		CreateDirectoryW(log_name, NULL);
+	}
+
+	ustrncat(log_name, L"kablam.txt", NUMBEROF(log_name));
+
+	*Memory::GetAddress<bool*>(0x0, 0x4904E8) = true;	// Set logging to true
```
Kantanomo: This should probably not be enforced as always on, probably should be a setting

Berthalamew: added -log parameter to enable it

PR #866, opened 2025-06-10
render changes and hud corruption fix when using the address sanitizer

Berthalamew:
- hud corruption issue caused since new_hud_widget_anchor_calculate_point should take a int32 instead of an int16 enum
- add scenery_datum structure
- fix illegal size warnings in inlined asm
- label scenery lightmap policy member and add scenery lightmap function
- add render visibility structures and functions
- remove DRAW_RENDER_LAYER macro, replace with render_scene_geometry as that's how it's originally setup (get's inlined on debug and release builds)
- add backup and restore member functions that get inlined

### DIFF
--- a/xlive/Blam/Engine/cseries/cseries.h
+++ b/xlive/Blam/Engine/cseries/cseries.h
@@ -78,8 +78,17 @@ extern bool g_catch_exceptions;
 // ADDR_SERVER: file offset in h2server.exe
 // TYPE: function
 // __VA_ARGS__: arguments for the function we want to invoke
-#define INVOKE_BY_TYPE(_addr_client, _addr_server, _type, ...) \
+// NOTE:
+// if server or client is equal to 0 then we pass the same address for both (ONLY ON RELEASE BUILDS)
+// REASON:
+// optimizes out a cmov or related instruction when building release whenever we call a function that doesn't have a dedi address specified
+#ifdef NDEBUG
+#define INVOKE_BY_TYPE(_addr_client, _addr_server, _type, ...)	\
+Memory::GetAddress<_type>(_addr_client == 0 ? _addr_server : _addr_client, _addr_server == 0 ? _addr_client : _addr_server)(__VA_ARGS__)
+#else
+#define INVOKE_BY_TYPE(_addr_client, _addr_server, _type, ...)	\
 Memory::GetAddress<_type>(_addr_client, _addr_server)(__VA_ARGS__)
+#endif
 
 #define INVOKE(_addr_client, _addr_server, _fn_decl, ...) INVOKE_BY_TYPE(_addr_client, _addr_server, decltype(_fn_decl)*, __VA_ARGS__)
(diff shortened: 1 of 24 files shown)


xlive/Blam/Engine/cseries/cseries.h
```diff
@@ -78,8 +78,17 @@ extern bool g_catch_exceptions;
...
 // TYPE: function
 // __VA_ARGS__: arguments for the function we want to invoke
-#define INVOKE_BY_TYPE(_addr_client, _addr_server, _type, ...) \
+// NOTE:
+// if server or client is equal to 0 then we pass the same address for both (ONLY ON RELEASE BUILDS)
+// REASON:
+// optimizes out a cmov or related instruction when building release whenever we call a function that doesn't have a dedi address specified
+#ifdef NDEBUG
+#define INVOKE_BY_TYPE(_addr_client, _addr_server, _type, ...)	\
+Memory::GetAddress<_type>(_addr_client == 0 ? _addr_server : _addr_client, _addr_server == 0 ? _addr_client : _addr_server)(__VA_ARGS__)
```
nukeulater, line 87: have you even measured the perf gain of this change? I honestly perceive it as pointless.

xlive/stdafx.cpp
```diff
@@ -7,4 +7,9 @@ std::random_device rd;
 extern "C" {
 	__declspec(dllexport) DWORD NvOptimusEnablement = 0x00000001;
 	__declspec(dllexport) DWORD AmdPowerXpressRequestHighPerformance = 0x00000001;
-}
\ No newline at end of file
+}
+
+extern "C" int LLVMFuzzerTestOneInput(uint8_t* buf, size_t len)
```
nukeulater: lol

xlive/Blam/Engine/cseries/cseries.h
Kantanomo: There's a better way to do this without chain inline ifs and relying onon the compiler to work it out And properly minimize the code generation. Make a constexpr function that checks the parameters and then calls the main function with the desired set of parameters.

Constexpr T* GetAddressConst<T>
{
  Constexpr If(server == 0 && client)
   Return GetAddress<T>(client);
else if(client == 0 && server)
   Return GetAddress<T>(server);
else
   Return GetAddress<T>(client, server);
}

Berthalamew: I haven't measured the performance impact but it saves 50KB to not slot the instructions in release_msvc.
I doubt it makes much of an impact performance wise but still better not to have the instructions in release.

xlive/stdafx.cpp
Berthalamew: Do you want this removed?
It doesn't do anything atm since the flag isn't passed but it lets you compile with the flag if anyone wants to test things using it.

Berthalamew: just removed it for now

PR #869, opened 2025-06-16
unit zoom rotation regression fix, game time and render updates

nukeulater:


xlive/Blam/Engine/shaders/shader_passes.h
```diff
@@ -149,7 +149,7 @@ enum e_shader_pass_implementation_flags : uint16
 	k_shader_pass_implementation_flag_count
 };
 
-enum e_shader_pass_texture_source_extern : int16
```
Berthalamew: this is an int16 in the tag definition

twist84: Maybe consider switching enums to `c_enum`

PR #877, opened 2025-07-07
rename s_player to player_datum

Berthalamew:
- remove member functions from player struct
- use s_player_identifier in the player struct instead of a uint64
- add unit_set_desired_grenade_type and unit_add_grenade_type_to_inventory
- add unit_try_and_get macro


Berthalamew:
Tested Gungame on a lan server to verify it works after changes

PR #879, opened 2025-07-09
Allow address reuse when creating sockets

Berthalamew:
- fixes a bug where multiple LAN servers would not show up in the server list
- make sure we only pass true if both sockets are created in XSocketManager::SystemLinkSocketInitialize
- add e_system_link_socket_type and update code to reference it
- update XSocketSend, XSocketSendTo, and XSocketRecv to return the proper values


nukeulater:
looks good but have it tested

loVis-fr:
‎ ‎ 
Tested with two PCs connected to the same local network, each running two server instances  
- All four appeared correctly in the Network Game Browser on both machines.
- Each instance could be joined from either side.
- Previously, each client could only see one instance from the host.
‎ ‎

PR #891, opened 2025-08-08
Downgrade Imgui version to old one

Himanshu-01:
* fixes winlator gamepad detection bug from upgrading imgui #778


nukeulater (changes requested):
I wouldn't revert it, this works on Windows therefore there's something wrong with the actual translation layer

nukeulater:
but if I were to look I'd look in the backend code (d3d9, win32)

xlive/3rdparty/imgui/backends/imgui_impl_win32.cpp
```diff
@@ -1,44 +1,38 @@
...
+#include "imgui_impl_win32.h"
+#ifndef WIN32_LEAN_AND_MEAN
+#define WIN32_LEAN_AND_MEAN
+#endif
+#include <windows.h>
+#include <tchar.h>
+#include <dwmapi.h>
+
+// Configuration flags to add in your imconfig.h file:
+#define IMGUI_IMPL_WIN32_DISABLE_GAMEPAD              // Disable gamepad support. This was meaningful before <1.81 but we now load XInput dynamically so the option is now less relevant.
```
nukeulater: have you tried messing with this?

Himanshu-01: seems to do the job, will use that instead

Berthalamew (approved):

PR #893, opened 2025-08-12
minor input and ui updates 

Himanshu-01:
* Add Splitscreen Mode into video settings to allow users to switch b/w horizontal & vertical modes
* Allow swapping grenades with 'A' in Bumper-Jumper layout

### DIFF
--- a/xlive/Blam/Engine/interface/screens/screen_splitscreen_setting.cpp
+++ b/xlive/Blam/Engine/interface/screens/screen_splitscreen_setting.cpp
@@ -4,6 +4,7 @@
 #include "interface/user_interface_memory.h"
 #include "interface/user_interface_screen_widget_definition.h"
 #include "interface/user_interface_utilities.h"
+#include <render/render.h>
 #include "tag_files/global_string_ids.h"
 #include "tag_files/tag_loader/tag_injection.h"
 #include "text/text_group.h"
@@ -101,8 +102,8 @@ static c_maximum_interface_text default_text;
 
 /* prototypes */
 
-static const e_splitscreen_list_items get_item_type_from_split_mode(e_override_splitscreen_mode mode);
-static const e_override_splitscreen_mode get_split_mode_from_item_type(e_splitscreen_list_items type);
+static const e_splitscreen_list_items get_item_type_from_split_mode(e_display_split_type mode);
+static const e_display_split_type get_split_mode_from_item_type(e_splitscreen_list_items type);
 
 /* public code */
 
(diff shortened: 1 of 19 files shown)


xlive/Blam/Engine/main/main_render.cpp
```diff
@@ -79,7 +81,15 @@ void __cdecl main_render_player_view(void)
 	else
 	{
 		const bool use_horizontal_split = g_debug_render_horizontal_splitscreen || rasterizer_get_display_type() == _display_type_4_by_3;
-		display_split_type = use_horizontal_split ? _display_split_type_horizontal : _display_split_type_vertical;
+		
+		if(H2Config_split_mode == e_override_splitscreen_mode::automatic)
```
Berthalamew: This code should go before the horizontal split, the global should override the Config setting if it's set

Berthalamew (changes requested):

xlive/Blam/Engine/main/main_render.cpp
Himanshu-01: should just remove that horizontal boolean imo

nukeulater (approved):

xlive/Blam/Engine/interface/screens/screen_splitscreen_setting.cpp
```diff
@@ -0,0 +1,287 @@
...
+		return _display_split_type_none;
+		break;
+	case _item_vertical:
+		return _display_split_type_vertical;
+		break;
+	case _item_horizontal:
+		return _display_split_type_horizontal;
+		break;
+	}
+	return _display_split_type_none;
```
Berthalamew: use a result variable instead of using a return for every case.

xlive/Blam/Engine/interface/screens/screen_splitscreen_setting.h
```diff
@@ -0,0 +1,60 @@
+#pragma once
+#include "interface/user_interface.h"
+#include "interface/user_interface_widget_list.h"
+#include "interface/user_interface_widget_list_item.h"
+#include "interface/user_interface_widget_window.h"
+#include "main/game_preferences.h"
+
+/* constants */
+/* enums */
+enum
```
Berthalamew, line 8: This should go under constants

xlive/Blam/Engine/interface/screens/screen_video_settings.h
```diff
@@ -13,7 +13,7 @@
 
 enum
 {
-	k_no_of_visible_items_for_video_settings = (8 + 1)
+	k_no_of_visible_items_for_video_settings = (8 + 2)
```
Berthalamew, line 16: replace this with `k_no_of_visible_items_for_video_settings = k_total_no_of_video_settings_list_items` ?

xlive/H2MOD/Modules/Shell/Config.cpp
```diff
@@ -8,6 +8,7 @@
 #include "render/render_lod_new.h"
 #include "rasterizer/dx9/rasterizer_dx9.h"
 #include "rasterizer/dx9/rasterizer_dx9_9on12.h"
+#include <render/render.h>
```
Berthalamew: #include <render/render.h> -> #include "render/render.h"

xlive/H2MOD/Modules/Shell/Config.cpp
```diff
@@ -494,6 +496,7 @@ void SaveH2Config()
 			CONFIG_SET(&ini, "use_d3d9on12", &g_rasterizer_dx9on12_enabled);
 			CONFIG_SET(&ini, "disable_amd_or_ati_patches", &g_rasterizer_dx9_driver_globals.disable_amd_or_ati_patches);
 			CONFIG_SET(&ini, "use_vsync", &H2Config_use_vsync);
+			CONFIG_SET(&ini, "splitscreen_mode", std::to_string(H2Config_split_mode).c_str());
```
Berthalamew: why is this being converted to a string?
Shouldn't we just cast this to an integer and then use the set functionality for that?

Berthalamew (changes requested):

xlive/Blam/Engine/interface/screens/screen_video_settings.h
Himanshu-01: nah , it might be same for this context but they both symbolize different things, better to keep them seperate

Berthalamew: alright

Berthalamew (approved):

PR #902, opened 2025-09-13
preliminary changes for terminal

Berthalamew:


xlive/Blam/Engine/physics/character_physics_mode_melee.cpp
```diff
@@ -182,8 +182,7 @@ void __thiscall c_character_physics_mode_melee_datum::update_internal
 				m_melee_start_origin = *player_origin;
 				m_has_target = true;
 				m_flags = 0;
-				m_maximum_counter = blam_ticks_real_to_integer((real32)(m_weapon_is_sword ? 7 : 1) + lunge_tick_count_2);
-				// m_maximum_counter = (real32)(m_weapon_is_sword ? 7 : 1) + lunge_tick_count_2;
+				m_maximum_counter = (int32)((m_weapon_is_sword ? 7.f : 1.f) + lunge_tick_count_2);
```
nukeulater, line 185: why this change?

Berthalamew: Original code is converting from an int to a float and then back to an int again for the counter.
Makes more sense to just store them as floats and avoid the initial conversion

PR #916, opened 2025-10-07
move other applicable external libraries to separate projects

Berthalamew:
- allows us to build external libs in parallel before we block for the precompiled header for the project


.github/workflows/build.yaml
```diff
@@ -66,7 +66,6 @@ jobs:
     - name: Verify DXSDK installation
       run: |
         echo "DXSDK_DIR is ${{needs.fetch-dxsdk2010.outputs.dxsdk-path}}"
-        dir ${{needs.fetch-dxsdk2010.outputs.dxsdk-path}}
```
nukeulater, line 69: Should at least replace it with something

nukeulater (changes requested):

.github/workflows/build.yaml
Berthalamew: changed so we verify the folder exists now

PR #923, opened 2025-10-18
logger updates

nukeulater:
- move logging code out of xliveless
- export just ordinals when compiling in release mode, include function names only in debug
- removed hired gun main loop hacks


xlive/Blam/Engine/render/render_submit.cpp
```diff
@@ -62,9 +63,9 @@ void __cdecl render_submit_transparent_geometry(s_model_group_submit_data* model
 	}
 
 
-	for (int8 i = 0; i < model_data->count; ++i)
```
Berthalamew, line 65: Why change the datatype here?

nukeulater: if you verify the assembly it does the comparison in 32 bit

Berthalamew (approved):

PR #933, opened 2025-10-27
network event log condition fix and added SSE code path for cross_product3d

Berthalamew:
- removed exemption for warning 4189
- network event code used the incorrect condition check for other log levels

### DIFF
--- a/xlive/Blam/Engine/cseries/cseries.h
+++ b/xlive/Blam/Engine/cseries/cseries.h
@@ -158,10 +158,14 @@ ASSERT_EXCEPTION(STATEMENT, true);	\
 (void)0
 
 #define vassert(STATEMENT, FORMAT, ...)	\
-if (!(STATEMENT))						\
-{									\
-	DISPLAY_ASSERT(csprintf(g_temporary, NUMBEROF(g_temporary), (FORMAT), __VA_ARGS__));	\
-}									\
+do																								\
+{																								\
+	if (!(STATEMENT))																			\
+	{																							\
+		DISPLAY_ASSERT(csprintf(g_temporary, NUMBEROF(g_temporary), (FORMAT), __VA_ARGS__));	\
+	}																							\
+}																								\
+while(0)
 
 #define unreachable() DISPLAY_ASSERT("unreachable")
 
--- a/xlive/Blam/Engine/cseries/cseries.h
+++ b/xlive/Blam/Engine/cseries/cseries.h
@@ -157,6 +157,12 @@ if (!(STATEMENT))                                               \
 ASSERT_EXCEPTION(STATEMENT, true);	\
 (void)0
 
+#define vassert(STATEMENT, FORMAT, ...)	\
+if (!(STATEMENT))						\
+{									\
+	DISPLAY_ASSERT(csprintf(g_temporary, NUMBEROF(g_temporary), (FORMAT), __VA_ARGS__));	\
+}									\
+
 #define unreachable() DISPLAY_ASSERT("unreachable")
 
 #define halt() DISPLAY_ASSERT("halt()")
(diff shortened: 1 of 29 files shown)


xlive/Blam/Engine/cseries/cseries.h
```diff
@@ -157,6 +157,12 @@ if (!(STATEMENT))                                               \
 ASSERT_EXCEPTION(STATEMENT, true);	\
 (void)0
 
+#define vassert(STATEMENT, FORMAT, ...)	\
+if (!(STATEMENT))						\
```
nukeulater: why not wrap this in a do while (0)?

xlive/Blam/Engine/math/real_math.h
```diff
@@ -471,103 +534,56 @@ void vector3d_from_euler_angles2d(real_vector3d* forward, const real_euler_angle
...
-if (!valid_real_vector3d_axes2(forward, up))			\
-{														\
-	const char* string = csprintf(g_temporary, NUMBEROF(g_temporary), "%s, %s: assert_valid_real_vector3d_axes2(%f, %f, %f / %f, %f, %f)", #forward, #up, (*forward).i, (*forward).j, (*forward).k, (*up).i, (*up).j, (*up).k);	\
-	DISPLAY_ASSERT(string);								\
-}														\
+vassert(valid_real_vector3d_axes2(forward, up), "%s, %s: assert_valid_real_vector3d_axes2(%f, %f, %f / %f, %f, %f)", #forward, #up, (*forward).i, (*forward).j, (*forward).k, (*up).i, (*up).j, (*up).k);	\
 (void)0
 
 #define assert_valid_real_matrix4x3(matrix)		\
 if (!valid_real_matrix4x3(matrix))				\
```
nukeulater: same here, these macros are asking to be misused

xlive/Blam/Engine/rasterizer/dx9/rasterizer_dx9_shader_submit_new.cpp
```diff
@@ -144,161 +108,161 @@ void c_shader_submission_interface_new::stage_texture(
...
-			rasterizer_dx9_set_target_as_texture(stage, _rasterizer_target_water_refraction);
-			rasterizer_target_get_resolution(_rasterizer_target_water_refraction, res_x, res_y);
-			continue_staging = false;
-			break;
-		case _shader_pass_texture_source_extern_global_aux1:
+			rasterizer_dx9_set_texture(stage, _bitmap_type_2d, bitmap_usage_default, NONE, 0, 0.f);
+		}
+		else if (light_def->flags.test(_light_definition_texture_camera_gel))
+		{
+			stage_texture(stage, _shader_pass_texture_source_extern_global_texture_camera, res_x, res_y);
```
nukeulater, line 115: would probably be to just copy the code from `_shader_pass_texture_source_extern_global_texture_camera` case instead of recursively call the function

nukeulater: in the compiled code the compiler deduced the code is the same for each case, thus it added a jump to that code block instead of copying twice, but in src is likely duplicated rather than a recursive call

Berthalamew: Confirmed that this was a recursive call.

xlive/Blam/Engine/math/real_math.h
Berthalamew: changed

xlive/Blam/Engine/cseries/cseries.h
Berthalamew: changed

nukeulater (approved):

PR #937, opened 2025-11-09
add in-game console and hs compiler logic

Berthalamew:
- adds logic to run commands at startup using init.txt

image


xlive/Blam/Engine/math/integer_math.h
```diff
@@ -39,6 +39,35 @@ inline int16 rectangle2d_height(const rectangle2d* rect)
...
+	rectangle->y0 = y0;
+	rectangle->x1 = x1;
+	rectangle->y1 = y1;
+	return rectangle;
+}
+
+inline rectangle2d* offset_rectangle2d(rectangle2d* rectangle, int16 dx, int16 dy)
+{
+	rectangle->left += dx;
+	rectangle->left += dx;
```
nukeulater, line 53: Probably unintended duplicate line

xlive/Blam/Engine/math/real_quantization.cpp
```diff
@@ -68,13 +71,13 @@ int32 quantize_unit_vector(const real_vector3d* u)
 	return type | ((quantized_f1 | (quantize_real(f2, -1.0f, 1.0f, 8, true) << 8)) << 3);
 }
 
-void dequantize_unit_vector(int32 quantized_unit_vector, real_vector3d* out_unit_vector)
+void dequantize_unit_vector3d(int32 quantized_unit_vector, real_vector3d* out_unit_vector)
 {
-	real32 dequantized_f1 = dequantize_real((uint8)(quantized_unit_vector >>  3), -1.0f, 1.0f, 8, true);
-	real32 dequantized_f2 = dequantize_real((uint8)(quantized_unit_vector >> 11), -1.0f, 1.0f, 8, true);
+	real32 dequantized_f1 = dequantize_real((uint8)((quantized_unit_vector >>  3) & BYTE_MAX), -1.0f, 1.0f, 8, true);
```
nukeulater: Explain

xlive/H2MOD/Modules/MapManager/MapManager.h
```diff
@@ -21,7 +20,7 @@ class MapDownloadQuery {
 	bool ShouldStopDownload();
 	void StopDownload();
 
-	std::wstring m_clientMapFilenameWide;
+	wchar_t m_clientMapFilenameWide[32];
```
nukeulater: Do you guarantee a max of 32 character file names?

nukeulater: Also 32 instead of a defined constant...

xlive/Blam/Engine/physics/character_physics_mode_melee.cpp
```diff
@@ -180,8 +179,7 @@ void __thiscall c_character_physics_mode_melee_datum::update_internal
 
 				m_initial_position = *position;
 				m_has_target = true;
-				m_maximum_counter = blam_ticks_real_to_integer((real32)(m_weapon_is_sword ? 7 : 1) + lunge_tick_count_adjusted);
-				// m_maximum_counter = (real32)(m_weapon_is_sword ? 7 : 1) + lunge_tick_count_adjusted;
+				m_maximum_counter = (int32)((m_weapon_is_sword ? 7.f : 1.f) + lunge_tick_count_adjusted);
```
nukeulater, line 182: Heh.. 

Did you even check what blam_ticks_real_to_int() does?

nukeulater (changes requested):

xlive/Blam/Engine/hs/hs_runtime.cpp
```diff
@@ -0,0 +1,1311 @@
...
+
+static int32 hs_string_to_boolean(int32 n)
+{
+	const size_t length = cstrlen((const char*)n);
+	return hs_long_to_boolean(length);
+}
+
+static int32 hs_short_to_real(int32 s)
+{
+	const real32 v = (real32)(int16)s;
```
nukeulater, line 704: These don't really look right to me

xlive/Blam/Engine/hs/hs_runtime.cpp
```diff
@@ -0,0 +1,1311 @@
...
+
+static int32 hs_enum_to_real(int32 e)
+{
+	const real32 v = (real32)e;
+	return (int32)v;
+}
+
+static int32 hs_real_to_short(int32 r)
+{
+	return (int16)(int32)*(real32*)&r;
```
nukeulater, line 722: Forcefully reinterpreting a float to a 32 bit value then downcasting it to 16 bit, while the param is not even a float doesn't sound like the intended behavior, same with the other functions...

I'd double check again if the implementation is valid

xlive/Blam/Engine/cseries/cseries.h
```diff
@@ -124,41 +137,32 @@ static_assert (sizeof(STRUCT) == (_SIZE), "Invalid size for struct ("#STRUCT") e
...
-		__debugbreak();							\
-}												\
-(void)0
 
-#define DISPLAY_ASSERT_EXCEPTION(STATEMENT, IS_EXCEPTION)		\
-display_assert(STATEMENT, __FILE__, __LINE__, IS_EXCEPTION);	\
-ASSERT_TRIGGER_EXCEPTION(IS_EXCEPTION);							\
+#define DISPLAY_ASSERT_EXCEPTION(STATEMENT, FILE, LINE, IS_EXCEPTION)	\
+display_assert(STATEMENT, FILE, LINE, IS_EXCEPTION);			\
+if (is_debugger_present())												\
```
nukeulater: No do while(0) guards

nukeulater (changes requested):

xlive/Blam/Engine/hs/hs_runtime.cpp
twist84: Unfortunately with the way Bungie did callbacks for hs casting this is how it is

xlive/Blam/Engine/math/real_quantization.cpp
Berthalamew: Removes the upper bits of the vector before casting to a uint8, caused errors when using the compiler flag to break when a value was truncated due to type size. Should probably replace this with MAX.

xlive/Blam/Engine/physics/character_physics_mode_melee.cpp
Berthalamew: Yes, however the actual assembly and instructions used differ between the builds of the game.
This is probably just because they didn't append `.f` to the numbers in the ternary statement. Since they add the result to a float and then brought down to an int it needs to be truncated.

xlive/Blam/Engine/hs/hs_runtime.cpp
Berthalamew: This is correct

xlive/Blam/Engine/hs/hs_runtime.cpp
Berthalamew: What twister mentioned is correct

xlive/H2MOD/Modules/MapManager/MapManager.h
Berthalamew: Got the number wrong here, it was 48 and I yea I need to make a constant for it.

xlive/Blam/Engine/math/integer_math.h
Berthalamew: Yup, thanks for catching

xlive/H2MOD/Modules/MapManager/MapManager.h
Berthalamew: nevermind I was correct initially, they just have the hash appended to it as well, I'll need to adjust the game options struct too.

Berthalamew: updated

nukeulater: This is the filename on the disk iirc, not the internal one, which could have many more characters.

xlive/Blam/Engine/physics/character_physics_mode_melee.cpp
nukeulater: It should be ceil truncated, not floor truncated, which is the default behavior.

xlive/Blam/Engine/math/real_quantization.cpp
Berthalamew: MAX would break this, going to leave it so we're explicit that we want those 8 bits and we don't want to take anything above that into account.

Berthalamew: added comment in code to explain this

xlive/Blam/Engine/cseries/cseries.h
Berthalamew: changed

xlive/Blam/Engine/math/integer_math.h
Berthalamew: fixed

xlive/Blam/Engine/math/real_math.cpp
```diff
@@ -102,6 +102,11 @@ void __cdecl real_math_reset_precision(void)
 	return;
 }
 
+int32 real_to_long_round(real32 v)
+{
+	return (int32)v;
```
nukeulater, line 107: The standard rounding mode is rounding down, i.e floor or discarding the decimal places. 

We need to round nearest, which the previous code did (+-0.5), or using lrintf(), but only when the FPU is properly configured with the round control flag set to nearest. I hope I'll adjust some of the floating point math configuration soon.

nukeulater (approved):

Issue #938, opened 2025-11-14
Ultrawide issues


indianaliam1:
Game seems to run fine in ultrawide mostly, howeever some minor problems mainly letterboxing in cutscenes 

<img width="3440" height="1440" alt="Image" src="https://github.com/user-attachments/assets/e5be73bc-4cf6-48cd-8d19-8d4ea4ba6ca3" />

gameplay doesn't seem to have issues apart from hud anchoring to the corners, which is fine in my opinion but could cause visibility issues on wider displays (32:9 or multimonitor setups)
<img width="3440" height="1440" alt="Image" src="https://github.com/user-attachments/assets/4340a64e-55ce-450c-8552-f106d4844891" />

Berthalamew:
fixed in #940

PR #940, opened 2025-11-16
treat 16:9 as the default aspect ratio for special aspect ratios

Berthalamew:
- this removes black bars on ultrawide monitors since we treat them as 16:9


Berthalamew:
fixes #938

PR #943, opened 2025-11-24
real math code research, add old FPU code variants of some of the functions

nukeulater:
- re-implement core save, load functionality

### DIFF
--- a/xlive/Blam/Engine/math/real_math.h
+++ b/xlive/Blam/Engine/math/real_math.h
@@ -2,6 +2,11 @@
 
 /* constants */
 
+#define _REAL_MATH_USE_LEGACY_FPU_INSTRUCTIONS 1
+
+// faster than actual calls to trig functions
+#define _REAL_MATH_USE_LEGACY_TRIG_INSTRUCTIONS 1
+
 #define _real_epsilon 0.0001f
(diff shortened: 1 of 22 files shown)


xlive/Blam/Engine/math/real_math.h
```diff
@@ -2,6 +2,11 @@
 
 /* constants */
 
+#define _REAL_MATH_USE_LEGACY_FPU_INSTRUCTIONS 1
+
+// faster than actual calls to trig functions
+#define _REAL_MATH_USE_LEGACY_TRIG_INSTRUCTIONS 1
```
Berthalamew, line 8: Please remove these and the assembly as we discussed prior.
Leave the helper functions so they just use the floating point versions of the global clib math functions for now...

Berthalamew (changes requested):

Berthalamew:
Tested the core loading and saving functionality when swapping between maps, works correctly.

xlive/Blam/Engine/math/real_math.h
nukeulater: No, these will remain as they are, FPU instructions respect the precision configuration set using _controlfp/_control87,  which is configured by `real_math_initialize()` to _PC_54 (i.e double)

PR #949, opened 2025-12-05
limit hs core functions access given certain contexts

nukeulater:
- fix legacy input text issues

### DIFF
--- a/xlive/Blam/Engine/hs/hs.cpp
+++ b/xlive/Blam/Engine/hs/hs.cpp
@@ -41,9 +41,7 @@
 #include "scenario/scenario.h"
 #include "scenario/scenario_definitions.h"
 
-#ifdef TERMINAL_ENABLED
 #include "scenario/scenario_kill_trigger_volumes.h"
-#endif
 
 /* constants */
 
--- a/xlive/Blam/Engine/hs/hs.cpp
+++ b/xlive/Blam/Engine/hs/hs.cpp
@@ -25,14 +25,12 @@
 
 #include "game/game.h"
 
-#ifdef TERMINAL_ENABLED
 #include "interface/hud_definitions.h"
 #include "interface/hud_messaging.h"
 #include "interface/interface.h"
 #include "interface/terminal.h"
 #include "main/console.h"
 #include "math/random_math.h"
-#endif
 
 #ifdef HS_COMPILER_ENABLED
 #include "models/model_definitions.h"
@@ -390,7 +388,9 @@ int32 g_error_buffer_length = 0;
 void hs_apply_patches(void)
 {
 	// hook the print command to redirect the text to our console
+#if TERMINAL_ENABLED
 	PatchCall(Memory::GetAddress(0xE9E50, 0xE49AC), hs_print);
+#endif
 
 	WritePointer((uintptr_t)&get_game_systems()[55].initialize_proc, hs_initialize);
 	WritePointer((uintptr_t)&get_game_systems()[55].dispose_from_old_map_proc, hs_dispose_from_old_map);
@@ -864,15 +864,13 @@ int16 hs_tokens_enumerate(const char* substring, int32 type_flags, const char** 
 
 /* private code */
 
+#ifdef HS_COMPILER_ENABLED
+
 static void __cdecl hs_print(const char* text)
 {
-#ifdef TERMINAL_ENABLED 
 	terminal_printf(global_real_argb_green, text);
-#endif
 	return;
 }
-
-#ifdef HS_COMPILER_ENABLED
 
 static bool sort_by_found_index(const void* s1, const void* s2, const void* compare_data)
 {
--- a/xlive/Blam/Engine/hs/hs.cpp
+++ b/xlive/Blam/Engine/hs/hs.cpp
@@ -386,9 +386,7 @@ int32 g_error_buffer_length = 0;
 void hs_apply_patches(void)
 {
 	// hook the print command to redirect the text to our console
-#if TERMINAL_ENABLED
 	PatchCall(Memory::GetAddress(0xE9E50, 0xE49AC), hs_print);
-#endif
 
 	WritePointer((uintptr_t)&get_game_systems()[55].initialize_proc, hs_initialize);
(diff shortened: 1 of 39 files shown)


xlive/Blam/Engine/hs/hs_function.cpp
```diff
@@ -4894,6 +4897,8 @@ const hs_function_definition* hs_function_table[] =
 	&sleep_until_definition,
 	&wake_definition,
 	&inspect_definition,
+
+#if !defined(RELEASE_DLL) && RELEASE_DLL == false
```
Berthalamew: please remove the ifdef here, the script table isn't designed to have intentional holes in it.

xlive/Blam/Engine/main/console.h
```diff
@@ -24,4 +24,30 @@ void console_warning(const char* format, ...);
...
+
+inline void console_close(void) {}
+
+inline void console_clear(void) {}
+
+inline void console_printf(const char* format, ...) {}
+
+inline void console_warning(const char* format, ...) {}
+
+inline void console_update(real32 dt) {}
```
Berthalamew: I'd prefer if these were just macros in release

xlive/Blam/Engine/main/console.cpp
```diff
@@ -359,8 +357,15 @@ static void console_complete(void)
 {
 	const char* matching_items[256];
 
+#if !defined(RELEASE_DLL) && RELEASE_DLL == false
+	uint32 type_flags = (uint32)NONE;
+#else
+	// enumerate just function names
+	uint32 type_flags = FLAG(3) | FLAG(5);
+#endif
```
Berthalamew: Not really a fan of intentionally disabling this logic if we're in release, we're supposed to be able to write and execute hs in the terminal as well.

Berthalamew (changes requested):

xlive/Blam/Engine/main/console.cpp
nukeulater: it disables specific suggestion enumerations from the table? you can still execute them if you know the command

xlive/Blam/Engine/hs/hs_function.cpp
nukeulater: The table has big chunks of it missing right below, as far as I am concerned anything before, including `inspect_definition` shouldn't be moved around/disabled but I am open to suggestions, to disable most of the commands that shouldn't be executed in a release DLL.

xlive/Blam/Engine/main/console.h
nukeulater: done

xlive/Blam/Engine/main/console.cpp
nukeulater: also suggestions are quite cluttered with lots of commands that I am not entirely sure they can be executed, loaded from the actual cache?

Berthalamew: Everything that's autofilled can be used in someway in the terminal right now, some of the auto filled statements aren't traditional commands but they can be used with other syntax to run things.

xlive/Blam/Engine/hs/hs_function.cpp
Berthalamew: The table is also used for in-game script execution, you can't "disable" the functionality entirely since we will need this when executing scripts.

xlive/Blam/Engine/interface/terminal.cpp
```diff
@@ -315,12 +316,10 @@ void __cdecl terminal_draw(void)
 			for (int32 i = terminal_globals.newest_output_line_index; i != NONE && font_size - font_size_scaled > 0; i = line->older_line_index)
 			{
 				line = (output_line_datum*)datum_get(terminal_globals.output_lines, i);
-				real32 fade_progress = (4.f - line->fade_progress) / 1.f;
+				real32 fade_progress = PIN(4.f - line->fade_progress, 0.0f, 1.0f);
```
Berthalamew, line 319: Why was this changed?

xlive/Blam/Engine/memory/static_arrays.h
```diff
@@ -73,18 +73,11 @@ class c_static_stack
...
 
-
-
 template<typename t_type, typename t_storage_type, size_t k_count>
 class c_flags_no_init 
 {
 	static_assert(std::is_unsigned_v<t_storage_type>, "error: t_storage_type is not an unsigned integral type");
-
-public:
-	const t_storage_type get_unsafe(void) const
```
Berthalamew, line 84: Please leave the get_unsafe and set_unsafe functions in the class

Berthalamew (changes requested):

xlive/Blam/Engine/interface/terminal.cpp
nukeulater: simple 1 liner?

xlive/Blam/Engine/memory/static_arrays.h
nukeulater: Not sure why they were added in the first place. And I don't care if they were in the base game

xlive/Blam/Engine/interface/terminal.cpp
Berthalamew: Divide by 1 is missing but it's not really relevant. Original game does the divide there but that constant didn't seem to be connected to any hs globals.

xlive/Blam/Engine/memory/static_arrays.h
Berthalamew: It's in the class in the original game and I'd like to keep them in the original class.

xlive/Blam/Engine/hs/hs.cpp
```diff
@@ -864,16 +862,14 @@ int16 hs_tokens_enumerate(const char* substring, int32 type_flags, const char**
 
 /* private code */
 
+#ifdef HS_COMPILER_ENABLED
```
Berthalamew: printing from the terminal isn't a part of the HS compiler logic, doesn't need to be pruned when the compiler isn't built.

Berthalamew (changes requested):

Berthalamew (approved):

PR #950, opened 2025-12-06
fix bug where certain remapped keys had issues with text translation

Berthalamew:


xlive/Blam/Engine/interface/user_interface_text.cpp
```diff
@@ -74,6 +78,122 @@ bool __cdecl user_interface_parse_string(wchar_t* string, size_t max_length, cha
...
+			string->set(src.get_string());
+		}
+		break;
+	case _key_colon:
+		if (get_current_language() == _language_french)
+		{
+			user_interface_global_string_get(_string_id_key_dollar, &src);
+			string->set(src.get_string());
+		}
+		if (get_current_language() == _language_italian)
```
nukeulater: else if, they cannot be the same at the same time, and do it for the other instances.

Berthalamew: changed

nukeulater (approved):

PR #957, opened 2025-12-20
fix `object_get_region_information()` causing banshee wing lights to stay on when region is destroyed

nukeulater:
- **fix legacy issue where joining players could disconnect connected players during map load**


xlive/Blam/Engine/objects/objects.h
```diff
@@ -396,7 +396,7 @@ void object_get_region_information(
 	int32* region_count,
 	int8** region_permutation_indices,
 	int8** region_render_permutation_indices,
-	object_region_information** region_information);
+	int8** region_information);
```
Berthalamew, line 399: this param should be the `object_region_information**` type

xlive/Blam/Engine/objects/objects.cpp
```diff
@@ -904,31 +903,31 @@ void object_get_region_information(
...
 	int8** region_render_permutation_indices,
-	object_region_information** region_information)
+	int8** region_information)
 {
 	const object_datum* object = object_get(object_index);
 
 	ASSERT(region_count);
 
-	object_region_information* regions = (object_region_information*)object_header_block_get_with_count(
+	int8* regions_indices = (int8*)object_header_block_get_with_count(
```
Berthalamew, line 912: the header is allocated with the `object_region_information` type so that's what we need to get here

Berthalamew (changes requested):

xlive/Blam/Engine/objects/objects.cpp
Berthalamew: Actually, this should probably stay as the current type but named differently.
Since it's not allocated as just the region info but has 2 bytes at the start.

xlive/Blam/Engine/objects/objects.h
Berthalamew: will change at a later point

xlive/Blam/Engine/objects/objects.cpp
Berthalamew: will change at a later point

xlive/Blam/Engine/simulation/simulation_watcher.h
```diff
@@ -6,7 +6,6 @@ class c_simulation_watcher
 {
 public:
 	void* vtable;
-	uint8 gap_0[4];
```
Berthalamew, line 9: was this incorrect before?

nukeulater: oh yeah

Berthalamew (approved):

PR #958, opened 2025-12-21
bug fixes

Berthalamew:
- properly fix portable log creation
- fix internal marker logic


xlive/H2MOD/Modules/Shell/Startup/Startup.cpp
```diff
@@ -147,7 +147,15 @@ void log_file_name_prepare(const wchar_t* logFileName, c_static_wchar_string<MAX
 	folders.append(instance_string);
 
 	// Place logs in server folder when portable or dedi
-	path->set(g_h2_portable || is_dedi ? L"\\" : g_h2_appdata_local_path);
+	if (g_h2_portable || is_dedi)
```
Kantanomo, line 150: this doesnt end up being the same as it originally was?

Berthalamew: no, it's considered a different path

Kantanomo (approved):

PR #968, opened 2025-12-28
separate xlive project into different project and lib files

Berthalamew:


nukeulater:
It's fine by me but you should've kept any refactoring out of this PR, like the type changes unless they were really necessary

Berthalamew:
Pretty much all of the refactoring was necessary to split things up and avoid including things in places they shouldn't be.
Going to be opening other PR's for most of the code refactors and other additions that are needed for this so this PR should eventually just become the VS project changes.

PR #975, opened 2026-01-11
misc havok changes

Berthalamew:
~~- ragdoll increase causes crashes in havok as the game ends up running out of memory~~
- move #if in terminal code
- add some havok memory code
- fix condition in havok_can_allocate_space_for_instance_of_object_definition


nukeulater:
Any proof of this being the problem?

I've played 16/16 instant respawn matches and it never crashed

nukeulater:
Also, keep the player limit removal enabled, limit just the ragdoll count to a lower value

Berthalamew:
> Any proof of this being the problem?
> 
> I've played 16/16 instant respawn matches and it never crashed

All the Havok crashes that have been reported to me only happened after the ragdoll limit was increased.
The crash also happens in Havok code where the second overflow pool ends up being used. Bungie knew about this and talked about it in a GDC doc for h2 where once they went over the second overflow pool the game would crash.
The crashes usually happen on BTB servers when there's a lot of players and complex vehicle objects being updated.

Berthalamew:
> Also, keep the player limit removal enabled, limit just the ragdoll count to a lower value

It was disabled above 8 players as well since each active player that isn't in a ragdoll state also has to be dealt with by Havok as well.
I wouldn't be surprised if that during testing they found that in games above 8 players they'd go over the limit when interacting with vehicles and other objects.

nukeulater:
Log the havok memory usage in the crash report to confirm for the next patch, then re-add the limit if necessary.

It might as well be the object creation code that wasn't matching

Berthalamew:
Object creation code issues were introduced after this crash started occurring.
I know it's running out of memory because the crash happens when it tries accessing the overflow pool memory global which ends up as a null pointer.

nukeulater:
I'd prefer increasing this memory limit if possible, but to confirm we should add the memory usage to the log, which might be useful in other instances as well.

Berthalamew:
Given the state the game is in when the crashes occur the memory calculation code will say that nothing is free as the overflow pool is set to a null pointer.
I'd like to disable it as it causes crashes and I've received complaints about this for a while.

Berthalamew:
After doing some more investigation seems like there's something else odd going on, going to re-add the patch and keep this for adding some of the havok structs/research.

nukeulater (approved):

PR #978, opened 2026-01-25
update tag injector to close remaining handle after loading

Berthalamew:


xlive/H2MOD.cpp
```diff
@@ -522,6 +522,10 @@ static bool __cdecl OnMapLoad(s_game_options* options)
 
 		resetAfterMatch = true;
 	}
+
+	// Clear remaining handle open in the tag injector at the end of post loading injections
+	tag_injection_clear_active_map();
```
Kantanomo: This should probably go under everything just above the return

Berthalamew: moved

Kantanomo (approved):

PR #984, opened 2026-02-19
move macros to memory from hook

Berthalamew:


xlive/Util/Memory.h
```diff
@@ -11,6 +10,44 @@ enum
...
+#define INVOKE(_addr_client, _addr_server, _fn_decl, ...) INVOKE_BY_TYPE(_addr_client, _addr_server, decltype(_fn_decl)*, __VA_ARGS__)
+// ### TODO find better name
+#define INVOKE_TYPE(_addr_client, _addr_server, _type, ...) INVOKE_BY_TYPE(_addr_client, _addr_server, _type, __VA_ARGS__)
+
+#define INVOKE_VFPTR_FN(_get_table_fn, _index, _type, ...) \
+	(this->**_get_table_fn<_type>(_index))(__VA_ARGS__)
+
+#define INVOKE_CLASS_FN(classobj, functionPtr)  ((classobj)->*(functionPtr))
+
+#ifdef LLVM
```
Kantanomo, line 39: Shouldn't the hooking macros stay in the hook file?

Berthalamew: It was moved over so we didn't have to include windows headers in libs that use these macros.
If we include hook somewhere we need windows headers

Kantanomo (approved):

PR #992, opened 2026-04-04
networking and sim changes

Berthalamew:
- misc adjustments and additions in networking code
- make interpolation data use thread_local since it's contents are only ever accessed from a single thread

### DIFF
--- a/xlive/Blam/Engine/game/players.cpp
+++ b/xlive/Blam/Engine/game/players.cpp
@@ -5,8 +5,9 @@
 #include "game/game_engine.h"
 #include "game/game_globals.h"
 #include "interface/user_interface_controller.h"
-#include "networking/network_event.h"
 #include "networking/session/network_session.h"
+#include "networking/network_event.h"
+#include "networking/network_game_definitions.h"
 #include "saved_games/game_variant.h"
 #include "scenario/scenario.h"
 #include "scenario/scenario_definitions.h"
@@ -206,7 +207,7 @@ uint32 player_appearance_required_bits()
 	return 39;
 }
 
-void __cdecl player_configuration_validate_character_type(
+void player_configuration_validate_character_type(
 	s_player_configuration* configuration_data)
 {
 	// Campaign verification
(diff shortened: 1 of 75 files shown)


nukeulater (dismissed):

Kantanomo:
You fixed up game_engine_has_teams() in this but then manually check if the bit is set in multiple locations in this pr. Would it be better to switch to using the function in those locations?

Berthalamew:
From what I've looked, `game_engine_has_teams` isn't used in any of those spots in the base game.
They check the team bit manually.

xlive/Blam/Engine/game/players.cpp
```diff
@@ -588,37 +580,123 @@ int16 local_player_count(void)
...
-	// TODO: implement
+	csmemset(player_appearance, 0, sizeof(*player_appearance));
+
 	return;
 }
 
+char const* player_identifier_get_string(
+	s_player_identifier const* player_id)
+{
+	static char player_id_string[sizeof(*player_id)*3 + 1];
```
Kantanomo, line 649: sizeof(s_player_identifier) * 3 + 1?

xlive/Blam/Engine/game/players.cpp
```diff
@@ -588,37 +580,123 @@ int16 local_player_count(void)
...
+		);
+	}
+
+	return player_id_string;
+}
+
+char const* clan_identifier_get_string(
+	s_clan_identifier const* clan_id)
+{
+	static char clan_id_string[sizeof(*clan_id)*3 + 1];
```
Kantanomo, line 669: sizeof(s_clan_identifier) * 3 + 1?

Berthalamew: 2 characters for the hex number and 1 for the separator after every byte.

xlive/Blam/Engine/game/players.cpp
Berthalamew: 2 characters for the hex number and 1 for the separator after every byte.

Kantanomo: why use have it grab the pointer object and sizeof that and not just use the actual type.

Berthalamew: It resolves at compile time but it's better to get the size of variable itself rather than assuming the variable is a specific type.

Kantanomo: but the type is explicit in the variable itself

Berthalamew: Yea, but I'm referencing it by the size of the variable that we're interacting with rather than the type itself.

Kantanomo (approved):

PR #993, opened 2026-04-19
Fix assert window and add helper functions

Berthalamew:
- assertion window will now pop up instead of the crash window
- add misc functions
- make use of player_get macro
- add c_enum class


xlive/Blam/Engine/physics/collision_bsp_definition.h
```diff
@@ -49,7 +65,7 @@ struct collision_surface
 {
 	uint16 plane;
 	uint16 first_edge;
-	e_collision_surface_flags flags;
+	uint8 /*e_collision_surface_flags*/ flags;
```
Kantanomo, line 70: you redefined the enum but removed the typing, considering it's usage you could just set the enum type to uint8 and not tie it into c_flags.

Berthalamew: That enum doesn't technically have a name in the original game so I just left our name there as a reference. It was originally a uint8

Berthalamew: ended up updating the rest of the definitions in the latest commit

Kantanomo (approved):

PR #996, opened 2026-04-22
set user agent to cartographer in curl options

Berthalamew:


xlive/Util/curl-interface.cpp
```diff
@@ -14,6 +14,7 @@ CURL* curl_interface_init_no_verify()
 	{
 		curl_easy_setopt(curl, CURLOPT_SSL_VERIFYPEER, 0L);
 		curl_easy_setopt(curl, CURLOPT_SSL_VERIFYHOST, 0L);
+		curl_easy_setopt(curl, CURLOPT_USERAGENT, "Cartographer");
```
Kantanomo: just for the sake of it cause your doing that append the version string to it, will probably never use it but why not.

Berthalamew: added

Kantanomo (approved):

PR #1002, opened 2026-06-19
matrix math changes

Berthalamew:
- fix incorrect logic in `matrix3x3_from_angles` and `matrix4x3_transform_vector`
- use correct variable names and parameter info

### DIFF
--- a/xlive/Blam/Engine/math/matrix_math.cpp
+++ b/xlive/Blam/Engine/math/matrix_math.cpp
@@ -1,7 +1,12 @@
 #include "stdafx.h"
 #include "matrix_math.h"
 
-real_matrix3x3* matrix3x3_from_forward_and_up(real_matrix3x3* matrix, const real_vector3d* forward, const real_vector3d* up)
+/* public code */
+
+real_matrix3x3* matrix3x3_from_forward_and_up(
+	real_matrix3x3* matrix,
+	real_vector3d const* forward,
+	real_vector3d const* up)
 {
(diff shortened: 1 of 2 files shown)


xlive/Blam/Engine/math/matrix_math.cpp
```diff
@@ -257,113 +312,167 @@ real_vector3d* matrix4x3_transform_vector(
...
+	real32 f = sine(yaw);
+	
+	real32 ad = a * d;
+	real32 bd = b * d;
 
-	matrix->scale = 1.0f;
-	matrix->vectors.forward.i = cosine_vector.i * cosine_vector.j;
-	matrix->vectors.forward.j = sine_vector.i * cosine_vector.k - sin_jk * cosine_vector.i;
-	matrix->vectors.forward.k = j_cos_k * cosine_vector.i + sine_vector.i * sine_vector.k;
+	matrix->scale = 1.f;
```
Kantanomo, line 349: should probably leave the scale unset by function, since the original doesn't and there is a posibility a usage of it has a scale set already. Same thing with the position fields

Berthalamew: not sure if you're looking at the wrong function but in my idb those fields are set

Kantanomo: was looking a 3x3 not 4x3

Kantanomo (approved):

PR #1012, opened 2026-07-31
rewrite light suppressor patch + hs changes

Berthalamew:
- rewrite light suppressor patch so we don't nop out the function to enable it
- change hs_inspect_boolean so we don't trigger an assert
- add new hs globals


xlive/Blam/Engine/render/render.cpp
```diff
@@ -187,11 +204,25 @@ uint32* global_frame_index_get(void)
...
 	return INVOKE(0x191032, 0x0, render_structure_find_camera, point, out_cluster_index, out_leaf_index);
 }
 
+void __cdecl render_light_suppressor_enable(void)
+{
+	// Don't enable the light suppressor if we've disabled it in the config
+	if (!H2Config_light_suppressor_disable)
+	{
+		*render_light_suppressor_enabled_get() = true;
+	}
```
Kantanomo: *render_light_suppressor_enabled_get() = !H2Config_light_suppressor_disable

Berthalamew: updated

Kantanomo (approved):

PR #1013, opened 2026-08-02
Halo playlist rewrite, new game settings ui and string injection

Kantanomo:


xlive/Blam/Engine/camera/following_camera.cpp
```diff
@@ -151,6 +158,14 @@ void __cdecl following_camera_player_control_get_camera_info(int32 user_index, s
 real32 following_camera_get_default_camera_field_of_view()
 {
 	s_saved_game_cartographer_player_profile* player_profile = cartographer_player_profile_get_by_user_index(updating_user_index);
+
+	s_game_variant* variant = get_game_variant();
+
+	if (game_is_multiplayer() && variant && variant->cartographer_settings.flags.test(_cartographer_variant_force_default_fov))
+	{
+		return k_following_camera_default_field_of_view_radians;
```
Berthalamew: Add a result variable and set it here instead of doing multiple returns

xlive/Blam/Engine/camera/observer.cpp
```diff
@@ -24,6 +26,10 @@ enum
 	OBSERVER_SIGNATURE = 'rad!'
 };
 
+
+constexpr real32 k_observer_default_field_of_view = 78.f;
```
Berthalamew, line 29: this could be stored as a enum constant instead of a global float

xlive/Blam/Engine/camera/following_camera.cpp
```diff
@@ -4,10 +4,17 @@
...
 #include "observer.h"
+#include "game/game.h"
 #include "game/game_globals.h"
 #include "game/player_control.h"
+#include "saved_games/game_variant.h"
 #include "saved_games/cartographer_player_profile/cartographer_player_profile.h"
 
+/* constant */
+
+constexpr real32 k_following_camera_default_field_of_view = 78.f;
```
Berthalamew, line 15: this could be stored as a enum constant instead of a global float

xlive/Blam/Engine/cutscene/cinematics.cpp
```diff
@@ -61,12 +61,6 @@ void __cdecl cinematic_update(void)
 	return;
 }
 
-void __cdecl draw_quad(const rectangle2d* points, pixel32 rect_color)
```
Berthalamew, line 64: This function is located in this .cpp file

xlive/Blam/Engine/game/game_engine.cpp
```diff
@@ -57,10 +57,6 @@ bool __cdecl game_engine_get_change_colors(s_player_appearance* player_profile,
 	return INVOKE(0x6E5C3, 0x6D1BF, game_engine_get_change_colors, player_profile, team_index, change_colors);
 }
 
-bool __cdecl game_engine_variant_cleanup(uint16* flags)
```
Berthalamew, line 60: This function is located in this .cpp file

xlive/Blam/Engine/game/game_engine_util.cpp
```diff
@@ -43,3 +43,30 @@ bool __cdecl sub_4701B6(datum player_index)
 {
 	return INVOKE(0x701B6, 0x0, sub_4701B6, player_index);
 }
+
+e_game_variant_description_index game_engine_type_get_variant_description_index(e_game_engine_type type)
+{
+	switch (type)
```
Berthalamew, line 50: add result variable so this can have a single return

xlive/Blam/Engine/game/game_time.cpp
```diff
@@ -197,6 +201,66 @@ real32 game_time_get_leftover(void)
...
+				game_time_globals->game_speed = 5.f;
+				break;
+			case _game_speed_modifier_none:
+			default:
+				game_time_globals->game_speed = k_game_time_default_game_speed;
+				break;
+		}
+	}
+	else
+		game_time_globals->game_speed = k_game_time_default_game_speed;
```
Berthalamew, line 260: add brackets

xlive/Blam/Engine/game/game_time.cpp
```diff
@@ -197,6 +201,66 @@ real32 game_time_get_leftover(void)
...
+
+	game_time_globals->initialized = true;
+}
+
+void game_time_apply_patches()
```
Berthalamew: this should be the first function definition in the cpp file and prototype in the header

xlive/Blam/Engine/game/game_time.h
```diff
@@ -1,5 +1,9 @@
 #pragma once
 
+/* constants */
+
+constexpr real32 k_game_time_default_game_speed = 1.f;
```
Berthalamew, line 5: this global should reside inside of the .cpp file and be an extern here

xlive/Blam/Engine/game/player_control.cpp
```diff
@@ -12,6 +13,12 @@
 #include "units/unit_definitions.h"
 
 #include "H2MOD/Modules/CustomVariantSettings/CustomVariantSettings.h"
+#include "saved_games/game_variant.h"
+
+/* constants */
+
+constexpr real32 k_player_default_field_of_view = 78.f;
```
Berthalamew, line 19: This can be an enum constant

xlive/Blam/Engine/game/players.cpp
```diff
@@ -743,3 +749,29 @@ static void player_configuration_validate_team(
 
 	return;
 }
+
+int32 player_get_spawn_protection_time(real32 timer)
```
Berthalamew: this needs to be static

xlive/Blam/Engine/game/players.cpp
```diff
@@ -743,3 +749,29 @@ static void player_configuration_validate_team(
...
+{
+	s_game_variant* variant = get_game_variant();
+
+	if (game_is_multiplayer() && variant)
+	{
+		switch (variant->cartographer_settings.spawn_protection)
+		{
+		case _player_spawn_protection_timer_none:
+			return 0;
```
Berthalamew: add result variable so we can do a single return

xlive/Blam/Engine/halo_playlist/definitions/halo_playlist_assault_property.cpp
```diff
@@ -0,0 +1,26 @@
...
+//    { L"Bomb Arm Time",                _halo_playlist_assault_property_bomb_arm_time },
+//    { L"Sticky Arming",                _halo_playlist_assault_property_sticky_arming },
+//    { L"Slow With Bomb",               _halo_playlist_assault_property_slow_with_bomb },
+//    { L"Bomb Hit Damage",              _halo_playlist_assault_property_bomb_hit_damage },
+//    { L"Damage Resistance With Bomb",  _halo_playlist_assault_property_damage_resistance_with_bomb },
+//    { L"Active Camo With Bomb",        _halo_playlist_assault_property_active_camo_with_bomb },
+//    { L"Vehicle Operation",            _halo_playlist_assault_property_vehicle_operation }
+//);
+
+e_halo_playlist_assault_property halo_playlist_item_collection_assault_property_get_value(wchar_t* value)
```
Berthalamew: wchar_t const* value

xlive/Blam/Engine/halo_playlist/definitions/halo_playlist_assault_property.h
```diff
@@ -0,0 +1,28 @@
+#pragma once
+#include "halo_playlist_item_collection.h"
+
+#define k_halo_playlist_assault_property_item_count 15
+
+enum e_halo_playlist_assault_property : int32
```
Berthalamew: this doesn't need to be defined as an int32

xlive/Blam/Engine/halo_playlist/definitions/halo_playlist_assault_property.h
```diff
@@ -0,0 +1,28 @@
+#pragma once
+#include "halo_playlist_item_collection.h"
+
+#define k_halo_playlist_assault_property_item_count 15
```
Berthalamew: enum constant

xlive/Blam/Engine/halo_playlist/definitions/halo_playlist_cartographer_setting.cpp
```diff
@@ -0,0 +1,184 @@
...
+    { L"None",          _player_spawn_protection_timer_none },
+    { L"3",             _player_spawn_protection_timer_three_seconds },
+    { L"Three Seconds", _player_spawn_protection_timer_three_seconds },
+    { L"5",             _player_spawn_protection_timer_five_seconds },
+    { L"Five Seconds",  _player_spawn_protection_timer_five_seconds },
+    { L"10",            _player_spawn_protection_timer_ten_seconds },
+    { L"Ten Seconds",   _player_spawn_protection_timer_ten_seconds }
+);
+
+wchar_t* halo_playlist_item_collection_cartographer_setting_get_name(e_halo_playlist_cartographer_setting_property_type value)
```
Berthalamew: wchar_t const* for all the below functions

xlive/Blam/Engine/halo_playlist/definitions/halo_playlist_cartographer_setting.cpp
```diff
@@ -0,0 +1,184 @@
...
+{
+    return halo_playlist_item_collection_cartographer_setting_flag_write_to_variant(value, variant, _cartographer_variant_disable_dub_shot);
+}
+
+bool halo_playlist_item_collection_cartographer_game_speed_write_to_variant(wchar_t* value, s_game_variant* variant)
+{
+    e_game_speed_modifier game_speed = halo_playlist_item_collection_cartographer_game_speed_get_value(value);
+
+    if (game_speed == k_game_speed_modifier_invalid)
+        return false;
```
Berthalamew: add result variables to functions below for single return

xlive/Blam/Engine/halo_playlist/definitions/halo_playlist_cartographer_setting.h
```diff
@@ -0,0 +1,69 @@
+#pragma once
+#include "halo_playlist_item_collection.h"
+
+#define k_halo_playlist_cartographer_setting_property_count 11
+#define k_halo_playlist_cartographer_engine_mode_count 6
+#define k_halo_playlist_cartographer_game_speed_count 5
+#define k_halo_playlist_cartographer_gravity_count 8
+#define k_halo_playlist_cartographer_spawn_protection_count 9
```
Berthalamew: these can be enum constants

xlive/Blam/Engine/halo_playlist/definitions/halo_playlist_cartographer_setting.h
```diff
@@ -0,0 +1,69 @@
...
+
+enum e_halo_playlist_cartographer_setting_property_type : int32
```
Berthalamew: these don't need to be manually specified as int32

xlive/Blam/Engine/halo_playlist/definitions/halo_playlist_header.h
```diff
@@ -0,0 +1,17 @@
...
+
+enum e_halo_playlist_header_type : int32
+{
+	_halo_playlist_header_playlist = 0,
+	_halo_playlist_header_variant = 1,
+	_halo_playlist_header_match = 2,
+
+	k_halo_playlist_header_count,
+
+	_halo_playlist_header_none = -1,
```
Berthalamew: NONE

xlive/Blam/Engine/halo_playlist/definitions/halo_playlist_header.h
```diff
@@ -0,0 +1,17 @@
+#pragma once
+#include "halo_playlist_item_collection.h"
+
+#define k_halo_playlist_header_item_count 5
+
+enum e_halo_playlist_header_type : int32
+{
+	_halo_playlist_header_playlist = 0,
+	_halo_playlist_header_variant = 1,
+	_halo_playlist_header_match = 2,
```
Berthalamew: only _halo_playlist_header_playlist should be defined as = 0

xlive/Blam/Engine/halo_playlist/definitions/halo_playlist_item_collection.cpp
```diff
@@ -0,0 +1,77 @@
...
+	switch (*value)
+	{
+	case L'0':
+	case L'F':
+	case L'N':
+	case L'f':
+	case L'n':
+	{
+		*out_result = false;
+		return true;
```
Berthalamew: result variable should be defined here so we don't have to do multiple returns

xlive/Blam/Engine/halo_playlist/definitions/halo_playlist_item_collection.h
```diff
@@ -0,0 +1,38 @@
...
+
+#define PLAYLIST_ITEM_COLLECTION(_name, _expected_count, ...)          \
+    static s_halo_playlist_item _name##_items[] = { __VA_ARGS__ };                    \
+    static_assert(ARRAYSIZE(_name##_items) == (_expected_count),                      \
+        #_name " out of sync with enum");                                             \
+    s_halo_playlist_item_collection _name = { _name##_items, ARRAYSIZE(_name##_items) }
+
+struct s_halo_playlist_item
+{
+	wchar_t* name;
```
Berthalamew: const?

xlive/Blam/Engine/halo_playlist/halo_playlist.cpp
```diff
@@ -0,0 +1,1200 @@
...
+    reader.m_match_count = 0;
+    reader.m_section_buffer_item_count = 0;
+    reader.m_current_reader_char_index = 0;
+    reader.m_current_reader_mode = _halo_playlist_reader_seek_mode_new_line;
+    reader.m_current_section_type = _halo_playlist_header_none;
+    reader.m_playlist_header_found = false;
+
+    if (file_handle)
+    {
+        wchar_t buffer[16384]{};
```
Berthalamew: I don't think this needs to be initialized to zero?

xlive/Blam/Engine/halo_playlist/halo_playlist.cpp
```diff
@@ -0,0 +1,1200 @@
...
+        result = playlist_container->playlist.match_variant_count != 0 ?
+            _halo_playlist_loading_result_success :
+            _halo_playlist_loading_result_empty_playlist;
+
+        fclose(file_handle);
+    }
+    else
+    {
+        if (GetLastError() == ERROR_FILE_NOT_FOUND)
+            result = _halo_playlist_loading_result_file_not_found;
```
Berthalamew: brackets

xlive/Blam/Engine/halo_playlist/halo_playlist.cpp
```diff
@@ -0,0 +1,1200 @@
...
+
+    this->m_current_header_file_line = this->m_current_reader_line;
+
+    e_halo_playlist_header_type header_type = halo_playlist_item_collection_get_header_type(this->m_section_header_buffer);
+    this->m_current_section_type = header_type;
+    this->m_section_buffer_item_count = 0;
+
+    switch (header_type)
+    {
+	    case _halo_playlist_header_playlist:
```
Berthalamew, line 251: brackets shouldn't be included in cases since we're not defining variables within the new scope

xlive/Blam/Engine/halo_playlist/halo_playlist.cpp
```diff
@@ -0,0 +1,1200 @@
...
+}
+
+void __declspec(noinline) c_halo_playlist_reader::process_current_property()
+{
+    s_halo_playlist_section_item* section = &this->m_section_items[this->m_section_buffer_item_count];
+
+    section->file_line = this->m_current_reader_line;
+    for (uint32 i = this->m_current_reader_char_index; i > 0; --i)
+    {
+        if (section->value_buffer[i] == L' ' || section->value_buffer[i] == L'\t')
```
Berthalamew: brackets

xlive/Blam/Engine/halo_playlist/halo_playlist.cpp
```diff
@@ -0,0 +1,1200 @@
...
+        }
+
+        if (!invalid
+            && wcscmp(variant_name, L"")
+            && (!base_variant_found || base_game_variant)
+            && (!game_type_found || game_type >= _game_variant_description_slayer)
+            )
+        {
+
+            s_game_variant new_variant{};
```
Berthalamew: this doesn't need to be zero-init

xlive/Blam/Engine/halo_playlist/halo_playlist.cpp
```diff
@@ -0,0 +1,1200 @@
...
+
+bool c_halo_playlist_reader::process_variant_territories_setting(s_halo_playlist_section_item* section_item, s_game_variant* variant)
+{
+    return INVOKE_TYPE(0, 0x121A8, bool(__thiscall*)(c_halo_playlist_reader*, wchar_t*, wchar_t*, uint32, s_game_variant*),
+        this, section_item->name_buffer, section_item->value_buffer, section_item->file_line, variant);
+}
+
+bool c_halo_playlist_reader::property_name_is_valid(wchar_t* property_name)
+{
+    if (halo_playlist_item_collection_get_variant_property_type(property_name) != k_halo_playlist_variant_property_invalid)
```
Berthalamew: if then else if instead of doing multiple returns for each if statement

xlive/Blam/Engine/halo_playlist/halo_playlist.cpp
```diff
@@ -0,0 +1,1200 @@
...
+    if (this->m_playlist->error_log_count == k_halo_playlist_maximum_errors) {
+        this->m_playlist->error_log_full = true;
+        return;
+    }
+
+    s_halo_playlist_error* error = &this->m_playlist->errors[this->m_playlist->error_log_count];
+    
+	error->error_type = error_type;
+    error->file_line = file_line;
+    wcsncpy_s(error->property_name, NUMBEROF(error->property_name), property_name, _TRUNCATE);
```
Berthalamew: ustrncpy

xlive/Blam/Engine/halo_playlist/halo_playlist.cpp
```diff
@@ -0,0 +1,1200 @@
...
+
+        if (!variant)
+        {
+            this->log_error(_halo_playlist_error_match_variant_not_found, match->variant_line_in_file);
+            continue;
+        }
+
+        int32 map_id = NONE;
+        char* map_path{};
+        s_secure_map_id custom_map_id{};
```
Berthalamew, line 1069: These don't need to be zero init

xlive/Blam/Engine/halo_playlist/halo_playlist.cpp
```diff
@@ -0,0 +1,1200 @@
...
+
+    if (left->error_type != right->error_type)
+        return left->error_type - right->error_type;
+
+    int32 result = _wcsicmp(left->property_name, right->property_name);
+    
+	if (result)
+        return result;
+
+    result = _wcsicmp(left->property_value, right->property_value);
```
Berthalamew: ustricmp

xlive/Blam/Engine/halo_playlist/halo_playlist.h
```diff
@@ -0,0 +1,183 @@
...
+	_halo_playlist_error_variant_base_variant_and_game_type_both_set,
+	_halo_playlist_error_match_variant_not_found,
+	_halo_playlist_error_match_map_not_found,
+	_halo_playlist_error_max_variants,
+	_halo_playlist_error_max_matches,
+};
+
+enum e_halo_playlist_reader_seek_mode
+{
+	_halo_playlist_reader_seek_mode_new_line = 0x0,
```
Berthalamew: only define _halo_playlist_reader_seek_mode_new_line and add an extra enum below this that's unused

xlive/Blam/Engine/halo_playlist/halo_playlist.h
```diff
@@ -0,0 +1,183 @@
+#pragma once
+#include "definitions/halo_playlist_header.h"
+#include "main/map_manager.h"
+#include "saved_games/game_variant.h"
+
+enum
```
Berthalamew, line 8: make sure sections are properly labeled here

xlive/Blam/Engine/input/input_windows.cpp
```diff
@@ -781,6 +782,11 @@ uint16 __cdecl input_key_msec_down(
 	return INVOKE(0x2F030, 0x0, input_key_msec_down, key);
 }
 
+bool input_windows_key_pressed(int16 key)
```
Berthalamew: This should be removed

xlive/Blam/Engine/render/render_cartographer_ingame_ui.cpp
```diff
@@ -265,6 +278,238 @@ void render_cartographer_update_message(const char* update_text, int64 update_si
...
 
+void render_cartographer_variant_settings()
+{
+	if (game_is_multiplayer())
+	{
+		key_stroke key;
+
+		input_peek_key(&key);
+
+		if (!input_windows_key_pressed(VK_F3))
```
Berthalamew: move checks and input to another file called `cartographer_variant_settings` in `interface` and add a global state that's checked here before rendering.

xlive/Blam/Engine/saved_games/game_variant.cpp
```diff
@@ -15,4 +16,51 @@ void __cdecl game_variant_build_default(s_game_variant* variant, e_game_variant_
 {
 	INVOKE(0x5B33D, 0x3CF9D, game_variant_build_default, variant, game_variant_type);
 	return;
+}
+
+s_game_variant* __cdecl get_default_game_variant_by_name(wchar_t* name)
+{
+	// Server only function?
+	if (Memory::g_memory_is_dedicated_server)
```
Berthalamew: If this is server only add an assert that makes sure only the server is calling it and just return the result of the INVOKE

xlive/Blam/Engine/saved_games/game_variant.cpp
```diff
@@ -15,4 +16,51 @@ void __cdecl game_variant_build_default(s_game_variant* variant, e_game_variant_
 {
 	INVOKE(0x5B33D, 0x3CF9D, game_variant_build_default, variant, game_variant_type);
 	return;
+}
+
+s_game_variant* __cdecl get_default_game_variant_by_name(wchar_t* name)
```
Berthalamew: wchar const*

xlive/Blam/Engine/cseries/cseries.h
```diff
@@ -306,3 +306,16 @@ __forceinline T* assert_return_internal(T* pointer)
 	return pointer;
 }
 #endif
+
+constexpr uint32 bits_required_for(uint64 max_value)
```
Berthalamew, line 310: you can just use log base 2 instead of this

xlive/H2MOD/Modules/OnScreenDebug/OnscreenDebug.cpp
```diff
@@ -17,7 +17,7 @@ std::recursive_mutex addTextMutex;
 
 void addDebugTextInternal(char* text) {
 
-	ASSERT(initialisedDebugText == true && "addDebugText() called before initialize");
+	//ASSERT(initialisedDebugText == true && "addDebugText() called before initialize");
```
Berthalamew: remove comment?

xlive/version.h
```diff
@@ -5,8 +5,8 @@
 // OR YOU CAN SET THE COMPATIBLE VERSION HIGHER IF YOU WANT TO ALLOW OLDER VERSIONS TO JOIN (EXECUTABLE_VERSION SHOULD NOT BE ABOVE IT)
 // ALSO THE VERSION SHOULD BE MODIFIED INCREMENTALLY, NOT RANDOMLY
 #if !defined(EXECUTABLE_VERSION) && !defined(COMPATIBLE_VERSION)
-#define EXECUTABLE_VERSION 11122
```
Berthalamew, line 8: undo changes

xlive/Project_Cartographer.vcxproj.user
```diff
@@ -6,7 +6,7 @@
     <LocalDebuggerWorkingDirectory>$(OutDir)</LocalDebuggerWorkingDirectory>
     <DebuggerFlavor>WindowsLocalDebugger</DebuggerFlavor>
     <LocalDebuggerCommandArguments>-nointro -windowed -novsync</LocalDebuggerCommandArguments>
-    <LocalDebuggerCommandArgumentsHistory>-nointro -windowed -novsync|</LocalDebuggerCommandArgumentsHistory>
+    <LocalDebuggerCommandArgumentsHistory>-live|-live -instance:asdf|-lan -instance:asdf|-windowed -novsync|-nointro -windowed -novsync|</LocalDebuggerCommandArgumentsHistory>
```
Berthalamew, line 9: undo changes

xlive/Blam/Engine/units/units.cpp
```diff
@@ -87,6 +87,16 @@ void __cdecl unit_get_camera_position(datum unit_index, real_point3d* out_point)
 	INVOKE(0x13D2EC, 0, unit_get_camera_position, unit_index, out_point);
 }
 
+void unit_get_aiming_vector(datum unit_index, real_vector3d* out_vector)
+{
+	unit_datum* unit = (unit_datum*)object_try_and_get_and_verify_type(unit_index, _object_mask_unit);
```
Berthalamew: unit_try_and_get

xlive/Blam/Engine/saved_games/game_variant.h
```diff
@@ -1,5 +1,8 @@
 #pragma once
 
+#define k_game_engine_oddball_maximum_balls 3
+#define k_multiplayer_variant_headhunter_parameter_count 7
```
Berthalamew: enum constants

Berthalamew (changes requested):

xlive/Blam/Engine/halo_playlist/halo_playlist.cpp
Kantanomo: if the file isn't large enough to fill the buffer, garbage data can be at the end of the buffer that causes the reader to fail

Berthalamew: Alright, can you make sure the fgetws call below uses NUMBEROF(buffer) instead of copying the buffer size again?

Berthalamew: Also, not sure if the while (!feof(file_handle)) should be at the top of the loop instead of the bottom incase the file starts with a EOF?

xlive/Blam/Engine/camera/following_camera.cpp
Kantanomo: its not a global float its a constexpr, which serves the same functionality as an enum constant but having the proper type.

xlive/Blam/Engine/game/players.cpp
```diff
@@ -743,3 +749,29 @@ static void player_configuration_validate_team(
...
+		case _player_spawn_protection_timer_one_second:
+			return game_seconds_to_ticks_round(1);
+		case _player_spawn_protection_timer_three_seconds:
+			return game_seconds_to_ticks_round(3);
+		case _player_spawn_protection_timer_five_seconds:
+			return game_seconds_to_ticks_round(5);
+		case _player_spawn_protection_timer_ten_seconds:
+			return game_seconds_to_ticks_round(10);
```
nukeulater: 10 seconds of spawn protection sounds insane

xlive/Blam/Engine/halo_playlist/definitions/halo_playlist_cartographer_setting.h
```diff
@@ -0,0 +1,69 @@
+#pragma once
+#include "halo_playlist_item_collection.h"
+
+#define k_halo_playlist_cartographer_setting_property_count 11
```
nukeulater: Isn't this just a redef of the count enum entry? 

(k_halo_playlist_cartographer_setting_property_count  -> k_halo_playlist_cartographer_setting_count)

xlive/Blam/Engine/halo_playlist/definitions/halo_playlist_oddball_property.h
```diff
@@ -0,0 +1,25 @@
+#pragma once
+#include "halo_playlist_item_collection.h"
+
+#define k_halo_playlist_oddball_property_item_count 12
```
nukeulater: same here (redef of the count property)

xlive/Blam/Engine/halo_playlist/halo_playlist.cpp
```diff
@@ -0,0 +1,1200 @@
...
+    if (file_handle)
+    {
+        wchar_t buffer[16384]{};
+        do
+        {
+            if (fgetws(buffer, 16384, file_handle))
+            {
+                reader.parse_file_section(buffer);
+            }
+            else if (!feof(file_handle))
```
nukeulater: maybe you meant else if (feof(file_handle))?

xlive/Blam/Engine/camera/following_camera.cpp
Berthalamew: It's a whole number so makes more sense to have it as an enum constant. Although either way a global isn't generated for it so either is fine.

xlive/Blam/Engine/halo_playlist/halo_playlist.cpp
```diff
@@ -0,0 +1,1200 @@
...
+            match->minimum_players = 0;
+
+        if (!match->maximum_players)
+            match->maximum_players = k_maximum_players;
+
+        ++this->m_match_count;
+    }
+}
+
+void __declspec(noinline) c_halo_playlist_reader::process_current_property()
```
nukeulater: noinline?

xlive/Blam/Engine/halo_playlist/halo_playlist.cpp
```diff
@@ -0,0 +1,1200 @@
...
+	        else
+	        {
+	            this->log_error(_halo_playlist_error_property_already_defined, this->m_current_reader_line, section->name_buffer);
+	            this->log_error(_halo_playlist_error_property_value_invalid, this->m_current_reader_line, section->name_buffer, section->value_buffer);
+	        }
+	        break;
+	    }
+	    case _halo_playlist_match_property_type_map:
+	    {
+	        if (!wcscmp(match->map, L""))
```
Berthalamew: ustrcmp

xlive/Blam/Engine/halo_playlist/halo_playlist.cpp
```diff
@@ -0,0 +1,1200 @@
...
+        uint32 variant_file_line = 0;
+
+        bool game_type_found = false;
+        e_game_variant_description_index game_type = k_game_variant_description_invalid;
+        uint32 game_type_line = 0;
+
+        bool base_variant_found = false;
+        wchar_t base_variant_name[32]{};
+        s_game_variant* base_game_variant{};
+        uint32 base_variant_file_line = 0;
```
Berthalamew, line 616: These 3 groups of variables should probably be their own structs

Berthalamew (changes requested):

xlive/Blam/Engine/camera/observer.h
```diff
@@ -9,6 +9,8 @@ enum
 	k_observer_command_values = 6
 };
 
+constexpr real_point2d k_observer_default_cross_hair_position{ {0.f, 0.138f} };
```
Berthalamew, line 12: values in cartographer_player_profile.cpp and AdvancedSettings.cpp should be replaced with a  reference to the y value in this

xlive/Blam/Engine/items/weapons.cpp
```diff
@@ -48,21 +53,95 @@ void weapon_barrel_idle(uint32 weapon_index, uint16 barrel_index)
...
+	if (available_rounds >= round_count)
+	{
+		int32 rounds_total = 0;
+		if (magazine_index >= 0 && magazine_index < weapon_def->weapon.magazines.count)
+		{
+			int32 max_rounds = weapon->weapon.magazines[magazine_index].rounds_inventory;
+
+			rounds_total = max_rounds;
+			if (round_count < max_rounds)
+				rounds_total = round_count;
```
nukeulater: rounds_total = min(round_count, max_rounds)

xlive/Blam/Engine/items/weapons.cpp
```diff
@@ -48,21 +53,95 @@ void weapon_barrel_idle(uint32 weapon_index, uint16 barrel_index)
...
+				{
+					weapon_datum* unit_weapon = weapon_get(owning_unit->unit.weapon_object_indices[index]);
+
+					if (unit_weapon && unit_weapon->definition_index == weapon->definition_index)
+					{
+						int32 max_rounds = unit_weapon->weapon.magazines[magazine_index].rounds_inventory;
+
+						int32 rounds_taken = round_count - rounds_total;
+
+						if (round_count - rounds_total > max_rounds)
```
nukeulater: same here, min()?

xlive/Blam/Engine/units/unit_action_system.cpp
```diff
@@ -0,0 +1,133 @@
...
+		popfd
+		popad
+
+		retn
+	}
+}
+
+/* public code */
+
+void unit_action_system_apply_patches()
```
Berthalamew: public code before private and private functions should all be static

xlive/H2MOD/Modules/CustomVariantSettings/CustomVariantSettings.cpp
```diff
@@ -236,12 +236,12 @@ namespace CustomVariantSettings
 
 	void Initialize()
 	{
-		ApplyHooks();
```
nukeulater, line 239: ?

xlive/H2MOD/Modules/CustomVariantSettings/CustomVariantSettings.h
```diff
@@ -33,7 +33,7 @@ namespace CustomVariantSettings
 
 		real32 ComputedGravity() const
 		{
-			return gravity * s_physics_constants::get_default_gravity();
```
nukeulater, line 36: remove if not necessary anymore

xlive/Blam/Engine/text/unicode.cpp
```diff
@@ -395,6 +395,13 @@ int32 utol(const wchar_t* string)
 	return _wtol(string);
 }
 
+bool __cdecl uniswcntrl(wchar_t* string)
```
Berthalamew: wchar_t const*

xlive/Blam/Engine/objects/damage.cpp
```diff
@@ -62,6 +63,12 @@ void __cdecl object_cause_damage(s_damage_data* damage_data, datum object_index,
 
 void __cdecl object_apply_damage_aftermath(datum object_index, s_damage_aftermath_data* aftermath_data)
 {
-	SET_BIT(aftermath_data->flags, _damage_aftermath_has_explosion_physics, currentVariantSettings.explosionPhysics);
-	INVOKE(0x17A25D, 0x0, object_apply_damage_aftermath, object_index, aftermath_data);
+	const s_game_variant* variant = get_game_variant();
+
+	if (game_is_multiplayer() && variant && variant->cartographer_settings.flags.test(_cartographer_variant_explosion_physics))
```
nukeulater, line 66: wasn't this implicitly multiplayer, being networked damage?

xlive/Blam/Engine/text/text_group.h
```diff
@@ -29,8 +55,8 @@ class c_language_pack
 	void string_list_get_normal_string(string_id id, c_maximum_interface_text* out_string, int32 strings_start_index, int32 strings_count);
 	void get_string_ids(string_id* array, int32 array_size, int32 starting_index, int32 max_count);
 
-
-private:
```
Berthalamew, line 33: members are being made public?

xlive/Blam/Engine/text/text_group.cpp
```diff
@@ -50,7 +59,14 @@ void c_language_pack::string_list_get_normal_string(
 	int32 strings_start_index,
 	int32 strings_count)
 {
-	INVOKE_TYPE(0x3E332, 0x0, void(__thiscall*)(c_language_pack*, string_id, c_maximum_interface_text*, int32, int32), this, id, out_string, strings_start_index, strings_count);
+	if (this->m_data_loaded)
+	{
+		c_maximum_interface_text string;
+		utf8* string_data = this->get_string_utf8(id, strings_start_index, strings_count);
+		utf8_string_to_wchar_string(string_data, (wchar_t*)&string, 512);
```
Berthalamew: ```utf8_string_to_wchar_string(string_data, string.get_buffer(), string.max_length());```

xlive/Blam/Engine/interface/multiplayer_variant_interface_cartographer_strings.h
```diff
@@ -0,0 +1,553 @@
+// ReSharper disable StringLiteralTypo
+
+#pragma once
+#include "multiplayer_variant_settings_interface_definition.h"
+#include "main/game_preferences.h"
+
+extern wchar_t* g_multiplayer_variant_interface_cartographer_title_strings[k_language_count][k_variant_setting_parameter_type_cartographer_count] =
```
nukeulater: I'm confused about the extern here even when you're defining the strings in the header?

xlive/Blam/Engine/text/text_group.cpp
```diff
@@ -41,7 +35,22 @@ utf8* c_language_pack::get_string_utf8(
 	int32 starting_index,
 	int32 max_count)
 {
-	return INVOKE_TYPE(0x3DEFD, 0x0, utf8*(__thiscall*)(c_language_pack*, string_id, int32, int32), this, id, starting_index, max_count);
+	int32 end_string_index = starting_index + max_count;
+
+	for (int32 str_index = starting_index; str_index < end_string_index; ++str_index)
+	{
+		if (!VALID_INDEX(str_index, this->m_num_of_strings))
```
Berthalamew, line 42: Shouldn't this check be moved into the for loop declaration itself?

xlive/Blam/Engine/tag_files/tag_loader/tag_injection_table.cpp
```diff
@@ -98,6 +109,18 @@ s_tag_injecting_table_entry* c_tag_injection_table::get_entry_by_injected_index(
...
 
+void c_tag_injection_table::get_entries_by_type(e_tag_group type, s_tag_injecting_table_entry* out_results)
+{
+	int32 out_index = 0;
+	for (uint16 i = 0; i < this->m_entry_count; ++i)
+		if (this->m_table[i].type.group == type)
+		{	
+			s_tag_injecting_table_entry* out = &out_results[out_index++];
+			s_tag_injecting_table_entry* in = &this->m_table[i];
+			csmemcpy(out, in, sizeof(s_tag_injecting_table_entry));
```
Berthalamew: *out = *in instead of a memcpy here

xlive/Blam/Engine/halo_playlist/halo_playlist.cpp
```diff
@@ -0,0 +1,1200 @@
...
+                }
+
+                if (base_variant_found)
+                    csmemcpy(&new_variant, base_game_variant, sizeof(s_game_variant));
+                else
+                    game_variant_build_default(&new_variant, game_type);
+            }
+
+            new_variant.flags &= ~1u;
+            wcsncpy_s(new_variant.variant_name, 32, variant_name, UINT_MAX);
```
nukeulater: _TRUNCATE, not UINT_MAX

xlive/Blam/Engine/render/render_cartographer_ingame_ui.cpp
```diff
@@ -265,6 +278,238 @@ void render_cartographer_update_message(const char* update_text, int64 update_si
...
+						}
+					}
+				}
+				else if (variant_categories[current_cell_index] == _variant_setting_category_type_cartographer_settings)
+				{
+					for (uint32 i = k_variant_setting_parameter_type_base_count + 1; i < k_variant_setting_parameter_type_base_count + 1 + k_variant_setting_parameter_type_cartographer_count; ++i)
+					{
+						e_variant_setting_parameter_type type = (e_variant_setting_parameter_type)i;
+
+						wchar_t title_buffer[512];
```
Berthalamew: these should probably be using c_maximum_interface_text and replace the printf calls below to the print functions of the static string class

xlive/Blam/Engine/halo_playlist/halo_playlist.cpp
Berthalamew: These should all be replaced with ustrncpy calls anyways

xlive/Blam/Engine/halo_playlist/halo_playlist.cpp
nukeulater: also else if, can't have 2 errors at the same time

xlive/Blam/Engine/render/render_cartographer_ingame_ui.cpp
```diff
@@ -265,6 +278,238 @@ void render_cartographer_update_message(const char* update_text, int64 update_si
...
+
+		const uint16 body_column_count = 3;
+		const uint16 body_row_count = 2;
+
+		const uint16 body_column_width = (body_rect.right - body_rect.left) / body_column_count;
+		const uint16 body_row_height = (body_rect.bottom - body_rect.top) / body_row_count;
+
+		int16 current_cell_index = 0;
+
+		const wchar_t* headers[body_column_count * body_row_count] = {
```
Berthalamew: These should be a static constant outside this function since the string array will be rebuilt every time the function is called

xlive/Blam/Engine/halo_playlist/definitions/halo_playlist_cartographer_setting.cpp
```diff
@@ -0,0 +1,184 @@
...
+    { L"50%",     _game_gravity_modifier_fifty_percent },
+    { L"75%",     _game_gravity_modifier_seventy_five_percent },
+    { L"125%",    _game_gravity_modifier_hundred_twenty_five_percent },
+    { L"150%",    _game_gravity_modifier_hundred_fifty_percent },
+    { L"175%",    _game_gravity_modifier_hundred_seventy_five_percent },
+    { L"200%",    _game_gravity_modifier_two_hundred }
+);
+
+PLAYLIST_ITEM_COLLECTION(g_halo_playlist_cartographer_spawn_protection_item_collection, k_halo_playlist_cartographer_spawn_protection_count,
+    { L"Default",       _player_spawn_protection_timer_one_second },
```
nukeulater, line 54: opinion, this should just allow spawn protection to be enabled/disabled, don't really see the point in having above 1 second spawn protection time periods

xlive/Blam/Engine/items/weapons.cpp
```diff
@@ -48,21 +53,95 @@ void weapon_barrel_idle(uint32 weapon_index, uint16 barrel_index)
...
+
+    if (game_is_multiplayer() && variant)
+       force_idle = barrel_def->damage_effect_reporting_type == _damage_reporting_type_battle_rifle && variant->cartographer_settings.flags.test(_cartographer_variant_disable_dub_shot);
+
+    if (!barrel_def->flags.test(_weapon_barrel_definition_dont_clear_fire_bit_after_recovering) || force_idle)
         weapon_barrel->flags.set(_weapon_barrel_fire_bit, false);
-    }
+}
+
+void weapon_take_inventory_rounds(datum weapon_index, int32 magazine_index, int32 round_count)
```
Berthalamew: needs to be static

xlive/Blam/Engine/interface/user_interface_globals_definition.h
```diff
@@ -0,0 +1,23 @@
+#pragma once
+#include "tag_files/tag_reference.h"
+
+/* structure */
+
+struct s_user_interface_tag_globals
```
Berthalamew: This struct is in user_interface_widget_group.h

xlive/Blam/Engine/halo_playlist/halo_playlist.cpp
```diff
@@ -0,0 +1,1200 @@
...
+    reader.m_playlist_header_found = false;
+
+    if (file_handle)
+    {
+        wchar_t buffer[16384]{};
+        do
+        {
+            if (fgetws(buffer, 16384, file_handle))
+            {
+                reader.parse_file_section(buffer);
```
nukeulater: maybe specify a len to this function, to avoid memset() to 0 of this buf?

xlive/Blam/Engine/halo_playlist/halo_playlist.h
```diff
@@ -0,0 +1,183 @@
...
+};
+ASSERT_STRUCT_SIZE(s_halo_playlist_container, 0x1E81C);
+
+class c_halo_playlist_reader
+{
+public:
+	s_halo_playlist* m_playlist;
+	s_halo_playlist_match m_matches[100];
+	uint32 m_match_count;
+	s_halo_playlist_section_item m_section_items[112];
```
Berthalamew: add enum constants for number of matches and match items?

xlive/Blam/Engine/halo_playlist/halo_playlist.h
```diff
@@ -0,0 +1,183 @@
...
+	s_halo_playlist_error errors[k_halo_playlist_maximum_errors];
+	uint32 error_log_count;
+	bool error_log_full;
+};
+ASSERT_STRUCT_SIZE(s_halo_playlist, 0x1E610);
+
+struct s_halo_playlist_loading_result
+{
+	int32 result_code;
+	wchar_t path[260];
```
Berthalamew: MAX_PATH

Berthalamew:
image

Not sure if we should remove the lock icon from the header or if it's something we're fine with keeping?

Berthalamew:
Also, should we move the anti cheat on message into this menu?
It should be easier in reports to see that along with all the game info at the same time.

Kantanomo:
> Also, should we move the anti cheat on message into this menu? It should be easier in reports to see that along with all the game info at the same time.

I think we should actually change the anti-cheat message entirely to be something that shows all the time to make our lives easier. My thought is just some icon or symbol that is in the same position as the current one and we just change the color to indicate if anti-cheat is running or not.

xlive/Blam/Engine/cseries/cseries.h
Kantanomo: that wouldn't produce the same output.

Berthalamew: well specifically the floor of logbase2 + 1

xlive/Blam/Engine/halo_playlist/definitions/halo_playlist_cartographer_setting.h
Kantanomo: the playlist item collections can have more items than the enum fits, if the same value gets mapped to multiple strings. ie ("10" = 10, "Ten" = 10) not all item collections do it, but I figured it'd be best to define them all the same rather than switching things around.

xlive/Blam/Engine/halo_playlist/halo_playlist.cpp
Kantanomo: if the file satarts with EOF the fgetws fails and goes into the else if that determines its not supposed to be the end of file and returns the unexpected end of file.

xlive/Blam/Engine/halo_playlist/halo_playlist.cpp
Kantanomo: no the !feof check is if the fgetws fails but eof hasn't been reached so its the unexpected end of file. Additionally it will look different from the original function as there was a bug in the original implementation that never actually closed the file handle it would just flush it.

xlive/Blam/Engine/objects/damage.cpp
Kantanomo: no it's not just limited to network events.

xlive/Blam/Engine/text/text_group.h
Kantanomo: members need to be accessible for string injection.

Berthalamew: Add member functions for operations that do string injection?

xlive/Blam/Engine/text/text_group.cpp
Kantanomo: ???

Berthalamew: ```for (int32 str_index = starting_index; str_index < end_string_index && VALID_INDEX(str_index, m_num_of_strings); ++str_index)```

Berthalamew: Or:
```for (int32 str_index = starting_index; str_index < end_string_index && str_index < m_num_of_strings; ++str_index)```
and then add an assert that makes sure starting_index is >= 0 and starting_index < m_num_of_strings

File: xlive/Blam/Engine/halo_playlist/halo_playlist.cpp (hunk 26 above)
Comment (Kantanomo) [reply xlive/Blam/Engine/halo_playlist/halo_playlist.cpp:251]: they are used for organization and readability it doesn't harm anything

File: xlive/Blam/Engine/interface/multiplayer_variant_interface_cartographer_strings.h (hunk 62 above)
Comment (Kantanomo) [reply xlive/Blam/Engine/interface/multiplayer_variant_interface_cartographer_strings.h]: they are made externs as I believe we determined that if a wchar_t* is made an extern it prevents duplications between TL's

Comment (Berthalamew) [reply xlive/Blam/Engine/interface/multiplayer_variant_interface_cartographer_strings.h]: These should probably be assigned inside of a .cpp file while the extern is here without the assignment.

File: xlive/Blam/Engine/halo_playlist/halo_playlist.cpp [hunk 74]
```diff
@@ -0,0 +1,1210 @@
...
+	            this->log_error(_halo_playlist_error_property_already_defined, this->m_current_reader_line, section->name_buffer);
+	            this->log_error(_halo_playlist_error_property_value_invalid, this->m_current_reader_line, section->name_buffer, section->value_buffer);
+	        }
+	        break;
+	    }
+	    case _halo_playlist_match_property_type_map:
+	    {
+	        if (!ustrcmp(match->map, L""))
+	        {
+	            if (!ustrcmp(section->value_buffer, L"") || !wcsncpy_s(match->map, NUMBEROF(match->map), section->value_buffer, UINT_MAX))
```
nukeulater: still a bunch of UINT_MAXs instead of _TRUNCATE

nukeulater:
also maybe a server MOTD in the game details window?

xlive/Blam/Engine/game/game_time.h
Kantanomo: its a const value making it an extern would be pointless other than placing it as a rdata member

xlive/Blam/Engine/cseries/cseries.h
Kantanomo: log2() cant be used in constexpr functions because it can't be evaluated at compile time

xlive/Blam/Engine/halo_playlist/halo_playlist.cpp
Kantanomo: ustrncpy wont work because it doesn't return the error_t that it's internal wcsncpy_s call uses, it just returns the pointer to the destination string.

xlive/Blam/Engine/game/game_time.h
Berthalamew: The value ends up in rdata regardless of if it's a constexpr or a global constant

xlive/Blam/Engine/halo_playlist/halo_playlist.cpp
Berthalamew: The result of that isn't used here though?

Kantanomo: not this specific one but others in the code

xlive/Blam/Engine/game/game_time.h
Kantanomo: constexpr variables get compiled away if they are not written to or referenced by a pointer which would be the case here.

xlive/Blam/Engine/halo_playlist/halo_playlist.cpp
Berthalamew: none of the calls that use that value would ever fail?
They take fixed size strings and copy them into another variable that's on the stack

xlive/Blam/Engine/game/game_time.h
Berthalamew: Yea, but the value is read in `game_time_setup_scenario` so it'll get put in rdata regardless in the compiled code. It's not used in a pre-compile step.

xlive/Blam/Engine/render/render_cartographer_ingame_variant_ui.cpp
```diff
@@ -0,0 +1,366 @@
...
+		}
+	}
+}
+
+/* private code */
+
+static real32 cartographer_ingame_variant_ui_fade_in_scale()
+{
+	real32 inv = 1.0f - g_cartographer_variant_settings_render_progress;
+	return 1.0f - (inv * inv);
```
nukeulater, line 336: why (inv * inv)?

Kantanomo: Instead of just scaling the alpha it uses a ease-out function

xlive/Blam/Engine/items/weapons.cpp
```diff
@@ -3,66 +3,50 @@
...
 /* public code */
 
 void weapons_apply_patches()
 {
-    PatchCall(Memory::GetAddress(0x15C60C), weapon_barrel_idle_usercall_to_rewritten);
-    PatchCall(Memory::GetAddress(0x1611AD), weapon_barrel_idle_usercall_to_rewritten);
-    PatchCall(Memory::GetAddress(0x162B38), weapon_barrel_idle_usercall_to_rewritten);
+	if (!shell_is_dedicated_server())
+	{
+		PatchCall(Memory::GetAddress(0x15C60C), weapon_barrel_idle_usercall_to_rewritten);
```
nukeulater: apply this on server as well, otherwise weapons might go out of sync

xlive/Blam/Engine/text/text_group.cpp
Kantanomo: that just makes the for loop messy and seems unnecessary, leave the loop logic in the loop and the filtering inside of it

xlive/Blam/Engine/halo_playlist/halo_playlist.cpp
Kantanomo: fgetws doesn't return a length so there isn't a clean way to get the proper length

xlive/Blam/Engine/halo_playlist/definitions/halo_playlist_cartographer_setting.cpp
Kantanomo: I think it'll be fun if people choose to use it for stupid game types

Kantanomo:
> also maybe a server MOTD in the game details window?

Possibly a future addition there is space that could be saved if we calculate the minimum cell height and add a footer. But it would also require wiring up changes to session parameters or some other network message to transmit the motd to the client

xlive/Blam/Engine/halo_playlist/halo_playlist.cpp
Kantanomo: seems unecessary to define a structure and setup initializer functions for it.

xlive/Blam/Engine/halo_playlist/halo_playlist.h
```diff
@@ -0,0 +1,194 @@
...
+	_halo_playlist_error_variant_base_variant_and_game_type_both_set,
+	_halo_playlist_error_match_variant_not_found,
+	_halo_playlist_error_match_map_not_found,
+	_halo_playlist_error_max_variants,
+	_halo_playlist_error_max_matches,
+};
+
+enum e_halo_playlist_reader_seek_mode
+{
+	_halo_playlist_reader_seek_mode_new_line,
```
Berthalamew: = 0 for the first entry of every enum

Berthalamew (changes requested):

xlive/Blam/Engine/halo_playlist/halo_playlist.cpp
```diff
@@ -0,0 +1,1198 @@
...
+	                {
+	                    this->log_error(_halo_playlist_error_property_value_invalid, base_variant_file_line, file_section->name_buffer, file_section->value_buffer);
+	                    valid_variant_setup = game_type_found == false;
+
+	                    if (!valid_variant_setup)
+	                        this->log_error(_halo_playlist_error_variant_base_variant_and_game_type_both_set, file_section->file_line);
+	                }
+	                else
+	                {
+	                    wcsncpy_s(base_variant_name, k_halo_playlist_maximum_text_length, file_section->value_buffer, _TRUNCATE);
```
Berthalamew: ustrncpy

xlive/Blam/Engine/halo_playlist/halo_playlist.cpp
Berthalamew: Just trying to think of ways to have it less cluttered but this is still fine

Berthalamew (changes requested):

xlive/Blam/Engine/halo_playlist/halo_playlist.cpp
Berthalamew: match_variant->variant = *variant

Berthalamew: match_variant->custom_map_id = custom_map_id

Berthalamew: ustricmp

Berthalamew: ustricmp

Berthalamew (changes requested):

xlive/Blam/Engine/interface/cartographer_variant_settings.h
```diff
@@ -0,0 +1,10 @@
+#pragma once
+
+/* globals */
+
+extern bool g_cartographer_variant_settings_render;
```
Berthalamew: make these two a struct and have a central global

Berthalamew (changes requested):

xlive/Blam/Engine/game/players.cpp
```diff
@@ -743,3 +749,31 @@ static void player_configuration_validate_team(
...
+			case _player_spawn_protection_timer_five_seconds:
+				result = game_seconds_to_ticks_round(5);
+			case _player_spawn_protection_timer_ten_seconds:
+				result = game_seconds_to_ticks_round(10);
+		}
+
+		return result;
+	}
+
+	return game_seconds_to_ticks_round(timer);
```
nukeulater: make this the single return path. also might be good to have a default in the switch which will use the time specified by the game

nukeulater (dismissed):

xlive/Blam/Engine/halo_playlist/definitions/halo_playlist_cartographer_setting.h
```diff
@@ -0,0 +1,81 @@
...
+	k_halo_playlist_cartographer_game_speed_count = 5,
+	k_halo_playlist_cartographer_gravity_count = 8,
+	k_halo_playlist_cartographer_spawn_protection_count = 9
+};
+
+/* enum */
+
+enum e_halo_playlist_cartographer_setting_property_type
+{
+	_halo_playlist_cartographer_setting_engine_mode,
```
Berthalamew, line 19: = 0

xlive/Blam/Engine/halo_playlist/definitions/halo_playlist_equipment_property.h
```diff
@@ -0,0 +1,30 @@
...
+#include "halo_playlist_item_collection.h"
+
+/* constants */
+
+enum
+{
+	k_halo_playlist_equipment_property_item_count = 8
+};
+
+/* globals */
```
Berthalamew, line 11: enums

Berthalamew (changes requested):

xlive/Blam/Engine/game/game_engine_util.cpp
```diff
@@ -43,3 +43,31 @@ bool __cdecl sub_4701B6(datum player_index)
...
 	return INVOKE(0x701B6, 0x0, sub_4701B6, player_index);
 }
+
+e_game_variant_description_index game_engine_type_get_variant_description_index(e_game_engine_type type)
+{
+	e_game_variant_description_index result = k_game_variant_description_invalid;
+	switch (type)
+	{
+	case _game_engine_type_ctf:
+		result = _game_variant_description_ctf;
```
Berthalamew: no break in each case statement

xlive/Blam/Engine/halo_playlist/definitions/halo_playlist_item_collection.h
```diff
@@ -0,0 +1,42 @@
+#pragma once
+#include "saved_games/game_variant.h"
```
Berthalamew, line 2: This shouldn't need to be here?

xlive/Blam/Engine/halo_playlist/definitions/halo_playlist_header.h
```diff
@@ -0,0 +1,26 @@
+#pragma once
+#include "halo_playlist_item_collection.h"
```
Berthalamew: include isn't needed here? (and the other definition headers)

xlive/Blam/Engine/input/input_windows.cpp
```diff
@@ -11,6 +11,7 @@
 #include "shell/shell_windows.h"
 
 #include "H2MOD/Modules/Shell/Config.h"
+#include "networking/network_event.h"
```
Berthalamew: not needed?

xlive/Blam/Engine/interface/interface.cpp
```diff
@@ -133,7 +133,7 @@ static void interface_splitscreen_render(void)
 			line.left = 0;
 			line.bottom = (resolution.y / 2) + line_size;
 			line.right = resolution.x;
-			draw_quad(&line, color);
```
Berthalamew, line 136: replace these with the original function

xlive/Blam/Engine/networking/network_game_definitions.cpp
```diff
@@ -0,0 +1,336 @@
...
+
+	packet->write_integer("starting-equipment-primary", variant->starting_equipment_primary, k_game_engine_starting_weapon_bits_required);
+
+	packet->write_integer("starting-equipment-secondary", variant->starting_equipment_secondary, k_game_engine_starting_weapon_bits_required);
+
+	switch (variant->variant_game_engine_index)
+	{
+	case _game_engine_type_ctf:
+	{
+		packet->write_integer("flags", variant->game_engine_variant.ctf.flags.get_unsafe(), k_ctf_engine_flags_bits_required);
```
Berthalamew, line 95: are we sure these aren't their own functions?

xlive/Blam/Engine/tag_files/tag_loader/tag_injection_manager.h
```diff
@@ -14,6 +15,16 @@ enum
 
 /* classes */
 
+struct s_tag_injection_string_container
```
Berthalamew, line 17: this is not a class

xlive/Blam/Engine/tag_files/tag_loader/tag_injection_manager.h
```diff
@@ -14,6 +15,16 @@ enum
...
+{
+	s_unicode_string_list_reference* unic_str;
+	s_string_reference* references;
```
Berthalamew: prefix `struct` so we don't have to include text_group.h in here

xlive/Blam/Engine/networking/network_game_definitions.cpp
```diff
@@ -0,0 +1,336 @@
...
+	case _game_engine_type_none:
+		return true; // native behavior??? why coop?
+
+	default:
+		return false;
+	}
+
+	// cartographer variant settings decoding
+
+	variant->cartographer_settings.version = (e_cartographer_variant_settings_version)packet->read_integer("cart-version", 32);
```
Berthalamew: use the enum version for this and only send the minimum bits required instead of 32

Berthalamew (changes requested):

xlive/Blam/Engine/halo_playlist/definitions/halo_playlist_item_collection.h
Kantanomo: it propogates the include to every playlist item collection without causing cyclical includes.

xlive/Blam/Engine/networking/network_game_definitions.cpp
Kantanomo: yes its one giant switch table

xlive/Blam/Engine/interface/interface.cpp
Berthalamew: draw_quad

Berthalamew (changes requested):

xlive/H2MOD.cpp
```diff
@@ -351,6 +351,19 @@ static bool __cdecl OnPlayerSpawn(datum player_index)
 
 	bool ret = p_player_spawn(player_index);
 
+	s_game_variant* variant = get_game_variant();
+
+	if (game_is_multiplayer() && variant && variant->cartographer_settings.flags.test(_cartographer_variant_invincible_players))
```
Berthalamew, line 356: add a comment specifying this is for the custom variant setting

xlive/Blam/Engine/units/unit_action_system.h
```diff
@@ -100,3 +122,6 @@ struct s_posture_definition
 };
 ASSERT_STRUCT_SIZE(s_posture_definition, 16);
 
+/* public functions */
```
Berthalamew: prototypes

xlive/Blam/Engine/units/unit_action_system.h
```diff
@@ -67,6 +67,15 @@ enum e_unit_action : int32
 	k_unit_action_count,
 };
 
+enum e_action_state_throw_state : int8
+{
+	_throw_idle = 0x0,
+	_throw_wind_up = 0x1,
+	_throw_in_hand = 0x2,
+	_throw_ending = 0x3,
+	k_throw_state_count = 0x4,
```
Berthalamew: _throw_idle = 0
and remove the assignment for the following values

xlive/Blam/Engine/units/unit_action_system.h
```diff
@@ -67,6 +67,15 @@ enum e_unit_action : int32
 	k_unit_action_count,
 };
 
+enum e_action_state_throw_state : int8
```
Berthalamew, line 70: remove `: int8` and either add a comment specifying the enum in the struct this is used in, or make it a c_enum

xlive/Blam/Engine/units/unit_action_system.cpp
```diff
@@ -0,0 +1,137 @@
...
+
+	object_get_markers_by_string_id(unit_index, _string_id_left_hand, &spawn_marker, 1);
+
+	damage_owner.owner_player_index = unit->unit.player_index;
+	damage_owner.owner_team_index = unit->unit.unit_team;
+	damage_owner.owner_object_index = unit_index;
+
+	object_placement_data_new(&placement_data, current_grenade->projectile.index, unit_index, &damage_owner);
+
+	SET_BIT(placement_data.flags, 2, true);
```
Berthalamew: enum name for 2?

xlive/Blam/Engine/units/unit_action_system.cpp
```diff
@@ -0,0 +1,137 @@
...
+	normalize3d(perpendicular3d(&placement_data.forward, &placement_data.up));
+
+	placement_data.position = spawn_marker.matrix.position;
+
+	datum new_object_index = object_new(&placement_data);
+
+	if (new_object_index == NONE)
+	{
+		action_state->throw_state = _throw_ending;
+		return;
```
Berthalamew: no need for return here

xlive/Blam/Engine/units/unit_action_system.cpp
```diff
@@ -0,0 +1,137 @@
...
+	SET_BIT(placement_data.flags, 2, true);
+
+	unit_get_aiming_vector(unit_index, &placement_data.forward);
+	normalize3d(perpendicular3d(&placement_data.forward, &placement_data.up));
+
+	placement_data.position = spawn_marker.matrix.position;
+
+	datum new_object_index = object_new(&placement_data);
+
+	if (new_object_index == NONE)
```
Berthalamew, line 125: This should be reversed so it's checking `new_object_index != NONE`

Berthalamew (changes requested):

xlive/H2MOD.cpp
Kantanomo: variant->cartographer_settings.flags.test

xlive/Blam/Engine/units/unit_action_system.h
Kantanomo: no

xlive/Blam/Engine/units/unit_action_system.cpp
Kantanomo: theres no reason to flip the logic around like that.

Berthalamew: Code is usually ordered that way where the correct condition is first then handling an error but either way works

Berthalamew (approved):

PR #1028, opened 2026-08-25
enable flag cloth position interpolation

nukeulater:
- freeze during the flag reset seems to be fixed, likely it was using garbage node position values from the stack because an interpolate function was reporting that the matrices were interpolated despite not writing to the node buffer because of the max interpolate distance check


nukeulater:
> * freeze during the flag reset seems to be fixed, likely it was using garbage node position values from the stack because an interpolate function was reporting that the matrices were interpolated despite not writing to the node buffer because of the max interpolate distance check

issue that has been fixed in a previous PR

Berthalamew (approved):
Tested on a LAN server with 2 game instances and works fine

PR #1045, opened 2026-09-20
absolute player index usage minor updates

nukeulater:


xlive/H2MOD/Variants/GunGame/GunGame.cpp
```diff
@@ -121,7 +121,6 @@ void GunGame::OnPlayerDeath(ExecTime execTime, datum player_index)
 
 void GunGame::OnPlayerSpawn(ExecTime execTime, datum player_index)
 {
-	const uint16 player_abs_index = DATUM_INDEX_TO_ABSOLUTE_INDEX(player_index);
```
Berthalamew, line 124: I think we should keep the abs index variable in cases where it's used more than once, rather than using `DATUM_INDEX_TO_ABSOLUTE_INDEX` multiple times in the function.

nukeulater: hopefully any other changes won't be necessary for the custom gametypes, I'll leave it this way

Berthalamew (approved):

